PR #101, opened 2022-03-03
[F] Use Headless-UI's modal

jen-castiron:
POC for using Headless UI's Dialog for a11y, scroll lock, and keyboard commands. Renders modals in react Portals. Was really easy to set up!


jen-castiron:
@dananjohnson  We can also try Radix's headless components if we want to use that library - the setup is very similar, so switching/testing should be easy.

PR #111, opened 2022-03-14
[F] Update button color, hover, and focus states

jen-castiron:
@dananjohnson I made a PR to call out some major updates to the button styling. Take a quick peep and let me know if something seems off 👀 

- Consolidated the button styles into styles/mixins/appearance.js
- Updated "secondary" style to be red, added "teacher" style (orange)
- Updated hover, focus states

### DIFF
--- a/components/atomic/Button/index.js
+++ b/components/atomic/Button/index.js
@@ -16,7 +16,7 @@ export default function Button({
       $isBlock={isBlock}
       $styleAs={styleAs}
       $hasIcon={!!icon}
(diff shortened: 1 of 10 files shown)


dananjohnson (commented):
The new buttons look good to me, @jen-castiron! My only real piece of feedback would be to use "educator" instead of "teacher" per our meeting last week. I'm going to update the auth modals as well.

jen-castiron:
@dananjohnson Updated & merged!

PR #112, opened 2022-03-16
[F] Authentication improvements

dananjohnson:


dananjohnson:
@jen-castiron would you mind testing out all the user journeys through authentication and seeing what you think? Thanks!

hooks/useAuthentication.js
```diff
@@ -130,97 +211,93 @@ export default function useAuthentication() {
...
+    const data = await authenticateFacebook({ code, pendingGroup });
 
     const returnRole =
-      role === "teacher" ? "googleSignInTeachers" : "googleSignInStudents";
-
-    if (data?.[returnRole]) {
-      setStateFromResponse(data[returnRole]);
-    }
+      pendingGroup === "teachers"
+        ? "facebookSignInTeachers"
```
jen-castiron, line 280: When I try to register as a teacher, I'm getting this error: 
```
"Cannot query field "facebookSignInTeachers" on type "Mutation". Did you mean "facebookSignInEditors", "facebookSignInEducators", or "facebookSignInStudents"?"
```
If the schema has changed to "Editors" then the google & facebook mutations will need to update too.

I also had trouble signing in as a student with facebook/google:
```
"No schema has been set for this user group"
```
Screen Shot 2022-03-16 at 1 31 51 PM

jen-castiron (commented):
@dananjohnson This feels a lot smoother, but there's a few little bugs. (If you're discussing them with Kyle right now, ignore my comments 😅 )

dananjohnson:
Thanks for reviewing, @jen-castiron! Yeah, I need to update everything from Teachers to Educators now that Kyle has pushed that change to the api repo. I'll do that and then merge.

That last error is from a plugin config issue in the backend. Seems like the schema was lost on a recent commit.

PR #125, opened 2022-03-28
[F] Add educator pages and page auth check

jen-castiron:
- Filters educator pages from global navigation
- Adds page auth check (all educator pages are restricted)
- _Does not_ prevent users from clicking on a restricted link (will be investigated in a follow up PR)


dananjohnson (approved):
Nice solution! 🙌🏻

PR #127, opened 2022-03-29
[F] Account dropdown and log out

jen-castiron:


jen-castiron:
@dananjohnson After exploring what other sites commonly do on logout, I found that they usually don't redirect the user unless they're on a restricted page. So I've updated this PR to only redirect if the user is on the "educatorPages" type. 👍

dananjohnson (changes requested):
Looking good, @jen-castiron! A couple little pieces of feedback:
* In the mobile menu, can the menu close when the user logs out?
* We need to announce the log-out event to SRs. Not sure how best to approach that. Maybe have a `<SRAnnouncement />` component somewhere on the page that subscribes to the authentication context? Or… ?

hooks/useBoundingBox.js
```diff
@@ -0,0 +1,20 @@
...
+
+  const set = () => {
+    const box = ref && ref.current ? ref.current.getBoundingClientRect() : {};
+    setBoundingBox(box);
+  };
+
+  useEffect(() => {
+    set();
+
+    window.addEventListener("resize", set);
```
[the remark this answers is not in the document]
jen-castiron: I'll debounce it! `useResizeObserver` isn't going to work, since the width and height of the element doesn't always change on browser width.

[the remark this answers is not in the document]
dananjohnson: Sounds good. You _can_ attach an observer to `window`, but I'm not really sure what you'd gain versus debouncing the `resize` event.

components/global/Header/Navigation.js
```diff
@@ -84,18 +91,33 @@ export default function Navigation({
         })}
         {theme === "mobile" && (
           <>
+            <li aria-live="polite">
+              <SRAuthStatus />
+            </li>
```
jen-castiron, line 96: Auth status added here for mobile

components/global/Header/UserNav/index.js
```diff
@@ -1,21 +1,66 @@
...
 
   return (
-    <Styled.Nav aria-label="User">
+    <Styled.Nav
+      aria-label="User"
+      aria-live="polite"
+      className="c-nav-list--desktop"
+      ref={ref}
+    >
+      <SRAuthStatus />
```
jen-castiron, line 23: @dananjohnson I wasn't quite sure where to put the auth status, since the location of the sign in / sign out buttons change on desktop and mobile. So I opted to add it near the buttons - here it is on the user nav. I might be over thinking this - would it be better to just have the status sit in the header?

dananjohnson: Maybe higher in the header? I could see us using it for other events too.

Btw, `role="status"` has implicit `aria-live="polite"`, so you don't need to add that to the parent nav. Unless there's some other intent I'm missing?

jen-castiron: Oh, sweet! I can leave the aria-live status around the user nav if we want to announce the button changes, but if we just want to announce the status I'll move it higher in the header.

dananjohnson: I think just announcing the status is fine. Announcing "Log in" after log out might be confusing without more context.

jen-castiron: @dananjohnson Cool cool. I've removed the extra aria-live regions and bumped the SRAuth announcement to the header. We can always extend it later to a more generic component if we need to. There's also the HOC (that I didn't see until now, of course) called `withLiveRegionChange`. If it looks good, approve it and I'll merge it in!

dananjohnson (approved):
Approved!

components/global/Header/UserNav/index.js
dananjohnson: Oh right, I forgot about the HOC. 🤷🏻 Sounds good though!

PR #128, opened 2022-03-31
[F] Create user account page

dananjohnson:


dananjohnson:
@jen-castiron I'd love a good review of this feature, which took a lot more effort than expected! You'll need to pull down the latest version of `develop` in the API branch first. Also, the slug for the account page will be whatever is defined in on the User Profile Page single in the backend. I haven't updated the navigation to get the correct URL yet, so you'll have to visit the page directly.

dananjohnson:
This PR should be complete and fully ready for review now!

components/modal/RegisterModal/RegisterForm/index.js
```diff
@@ -38,7 +38,6 @@ export default function RegisterForm({ onCancel }) {
       const errorSource = getErrorSource(invalid);
 
       if (errorSource === "email") {
-        console.info(`settings error on email: ${invalid.message}`);
         setError("email", { type: "manual", message: invalid.message });
       } else {
         setError("form", {
```
jen-castiron, line 43: I think this error is causing some unexpected problems. When there's an error on the form, I can't resubmit this form without refreshing the page:
Screen Shot 2022-04-05 at 12 54 00 PM

jen-castiron (commented):
I wasn't able to get this working locally (as we've been discussing in Slack), but the code looks OK to me. There's one area that I think was a bug before this PR, but would be great if it could get fixed too.

jen-castiron:
Couple more little things once I got the page working:
* Toggling the email subscription didn't activate the "Update profiles and preferences" button
* Can we use a teal background / white circle to show that this switch is on? The a11y is right, but it could use some extra color/contrast to help indicate that it's on. We can show the result to José for final approval. (Apologies if this was already brought up to José, I'm a little behind on that channel)
Screen Shot 2022-04-05 at 1 20 56 PM

* I ran into the same issue with the form not resetting in the password section. I added my current password, new password, and the button lit up so I clicked it. 🙃  Obviously the form isn't valid, but the error doesn't go away when I update any fields. If the error is set on a field name (rather than the form), it should reset when that field changes.
Screen Shot 2022-04-05 at 1 22 28 PM

dananjohnson:
Thanks for the feedback, @jen-castiron! I think I've addressed everything I could. Feel free to review once more or approve the PR if you think it's ready to go!

hooks/useAuthentication.js
```diff
@@ -188,30 +224,26 @@ export default function useAuthentication(data) {
...
+      data.authenticate?.user?.preferredLanguage
+    );
+    return handleSuccess(data.authenticate);
   }
 
   function signOut() {
     clearState();
 
-    if (typeHandle === "educatorPages") {
+    if (typeHandle === "educatorPages" || typeHandle === "userProfilePage") {
```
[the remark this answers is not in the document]
dananjohnson: Good idea! I've moved the logic to AuthorizePage. It'll watch for changes to `isAuthenticated` and redirect to the homepage if someone signs out.

dananjohnson:
I agree about the toggle too—let's ask José what he thinks about changing it at our next moot.

components/modal/RegisterModal/RegisterForm/index.js
dananjohnson: I think I've fixed all the issues around errors and additional form submissions.

jen-castiron (approved):
👍

PR #130, opened 2022-04-01
[F] Build and style Complex Table and Table Group content blocks

jen-castiron:
Requires @kylepratuch 's rubin-obs-api branch `kp-super-table` to be merged into develop.

### DIFF
--- a/components/content-blocks/ComplexTable/index.js
+++ b/components/content-blocks/ComplexTable/index.js
@@ -4,12 +4,17 @@ import * as Styled from "./styles";
 export default function ComplexTable({
   complexTable,
   plainText,
+  verticalAlignment,
   styleAs = "primary",
   isChild = false,
 }) {
   const renderTable = () => (
     <Styled.TableWrapper>
-      <Styled.Table as="table" $styleAs={styleAs}>
+      <Styled.Table
+        as="table"
+        $styleAs={styleAs}
+        $verticalAlignment={verticalAlignment}
+      >
         {plainText && <caption className="a-hidden">{plainText}</caption>}
         <tbody>
           {complexTable.map((row, i) => (
--- a/components/content-blocks/ComplexTable/index.js
+++ b/components/content-blocks/ComplexTable/index.js
@@ -10,21 +10,22 @@ export default function ComplexTable({
   const renderTable = () => (
     <Styled.TableWrapper>
       <Styled.Table as="table" $styleAs={styleAs}>
-        <caption className="a-hidden">{plainText}</caption>
-        {complexTable.map((row, i) => (
-          <Styled.TableRow key={i} as="tr">
-            {row.tableRow.map((cell, j) => (
-              <Styled.TableCell
-                as="td"
-                key={cell.id}
-                $row={i + 1}
-                $colWidth={cell.cellWidth}
-                $background={cell.cellBackground}
-                dangerouslySetInnerHTML={{ __html: cell.cellContent }}
-              />
-            ))}
-          </Styled.TableRow>
-        ))}
+        {plainText && <caption className="a-hidden">{plainText}</caption>}
+        <tbody>
+          {complexTable.map((row, i) => (
+            <Styled.TableRow key={i}>
+              {row.tableRow.map((cell, j) => (
+                <Styled.TableCell
+                  key={cell.id}
+                  $row={i + 1}
+                  colspan={cell.cellWidth}
+                  $background={cell.cellBackground}
+                  dangerouslySetInnerHTML={{ __html: cell.cellContent }}
+                />
+              ))}
+            </Styled.TableRow>
+          ))}
+        </tbody>
       </Styled.Table>
     </Styled.TableWrapper>
   );
--- a/components/content-blocks/ComplexTable/index.js
+++ b/components/content-blocks/ComplexTable/index.js
@@ -1,0 +1,37 @@
+import Container from "@/components/layout/Container";
+import * as Styled from "./styles";
+
+export default function ComplexTable({
+  complexTable,
+  plainText,
+  styleAs = "primary",
+  isChild = false,
+}) {
+  const renderTable = () => (
+    <Styled.TableWrapper>
+      <Styled.Table as="table" $styleAs={styleAs}>
+        <caption className="a-hidden">{plainText}</caption>
+        {complexTable.map((row, i) => (
+          <Styled.TableRow key={i} as="tr">
+            {row.tableRow.map((cell, j) => (
+              <Styled.TableCell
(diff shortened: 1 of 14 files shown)


components/content-blocks/ComplexTable/index.js
```diff
@@ -0,0 +1,37 @@
...
+      <Styled.Table as="table" $styleAs={styleAs}>
+        <caption className="a-hidden">{plainText}</caption>
+        {complexTable.map((row, i) => (
+          <Styled.TableRow key={i} as="tr">
+            {row.tableRow.map((cell, j) => (
+              <Styled.TableCell
+                as="td"
+                key={cell.id}
+                $row={i + 1}
+                $colWidth={cell.cellWidth}
```
jen-castiron, line 21: WIP - Need to use table cell's html prop for this

components/content-blocks/ComplexTable/styles.js
```diff
@@ -0,0 +1,56 @@
...
+          border-spacing: 5px;
+        `
+      : css`
+          --ComplexTable-border: 1px solid var(--black);
+          --ComplexTable-cell-bg: none;
+          border-collapse: collapse;
+        `}
+`;
+
+// export const Table = styled.div`
```
jen-castiron, line 26: WIP - Using grid would be nice, but required additional code to add missing cells, and the styling benefits were minimal.

components/content-blocks/TableGroup/styles.js
```diff
@@ -0,0 +1,28 @@
...
+  display: block;
+  flex: 1 1 auto;
+  background-color: var(--orange20);
+  border: 4px solid var(--orange20);
+  padding: 15px 32px;
+  min-height: ${fluidScale("90px", "70px")};
+  max-width: 373px;
+  font-weight: bold;
+  margin-inline-start: 20px;
+  margin-block-end: 10px;
```
jen-castiron, line 22: WIP - Add transitions here for button color/border states

lib/api/fragments/page.js
```diff
@@ -44,12 +44,13 @@ fragment pageFragmentFull on pages_pages_Entry {
         ${getImageFields("crop", 800, 600)}
       }
     }
-    subHeroText
-    subHeroHeader: plainText
-    subHeroColorScheme: colorScheme
+    # subHeroText
+    # subHeroHeader: plainText
+    # subHeroColorScheme: colorScheme
```
jen-castiron, line 49: WIP - remove this, KP's branch was behind develop, which has subheros on pages

jen-castiron:
@dananjohnson The tables are _so_ close to getting done! There's a weird bug though.
1. Add a Table Group content block to a page
2. Add a Complex Table to the Table Group (don't forget the title!)
3. Add some cells to the table

The rendered page will have an extra Complex Table under the Table Group. I couldn't figure out why the table is getting rendered twice... This query in GraphiQL shows the correct number of content blocks returned:
```
fragment complexTableBlock on contentBlocks_NeoField {
  ...on contentBlocks_complexTable_BlockType {
    id
    typeHandle
    plainText
    complexTable {
      ... on complexTable_BlockType {
        tableRow {
          ... on tableRow_tableCell_BlockType {
            id
            cellWidth
            cellBackground
            cellContent
          }
        }
      }
    }
  }
}

fragment tableGroupBlock on contentBlocks_NeoField {
  ...on contentBlocks_tableGroup_BlockType {
    id
    typeHandle
    items: children {
      ...complexTableBlock
    }
  }
}

query entry {
  entry(id:"7935") {
    ...on pages_pages_Entry {
      id
      title
      uri
      typeHandle
      title
      contentBlocks {
        ...complexTableBlock
        ...tableGroupBlock
      }
    }
  }
}
```
But the frontend renders an extra content block for each child table.

components/content-blocks/ComplexTable/index.js
```diff
@@ -0,0 +1,38 @@
...
+                  $background={cell.cellBackground}
+                  colSpan={cell.cellWidth}
+                  dangerouslySetInnerHTML={{ __html: cell.cellContent }}
+                />
+              ))}
+            </Styled.TableRow>
+          ))}
+        </tbody>
+      </Styled.Table>
+    </Styled.TableWrapper>
```
dananjohnson, line 30: I love how simple and straightforward this turned out to be!

components/content-blocks/TableGroup/index.js
```diff
@@ -0,0 +1,32 @@
+import Container from "@/components/layout/Container";
+import { Tab } from "@headlessui/react";
```
dananjohnson, line 2: Time for me to take a closer look at this library! 😁

dananjohnson (commented):
@jen-castiron looks really great! I tried reproducing one of the grayscale tables from the comps and noticed a few discrepancies/thought of a few questions. I'd love to hear your thoughts.

_Design_
Screen Shot 2022-04-05 at 9 35 21 AM

_Implementation_
Screen Shot 2022-04-05 at 9 48 15 AM

* little bit of scrollbar showing (because using `overflow: scroll` rather than `overflow: auto`?)
* should we allow vertical alignment to be defined to match comps? In the grayscale tables, it's to the baseline, but elsewhere it's centered
* should we allow column widths to be defined to match the comps?
* should we have the first row always be a `th`?
* for mobile, should we have a min-width for columns and let the table overflow scroll?

And I couldn't reproduce the issue with double tables rendering in the Table Group! I see just one. 🤷🏻 
Screen Shot 2022-04-05 at 9 55 33 AM

components/content-blocks/TableGroup/index.js
jen-castiron: Oh man, putting together the tabs was so simple with this library. ✨

jen-castiron:
@dananjohnson Thanks! I made some minor adjustments. 
* The column span can be set, but I don't think we want to allow the user to set a column width. I've set a min width on mobile, like you suggested. 
* I _think_ it's safe to assume all of the top row are going to be header cells... a checkbox could be added to the backend, but I'm not seeing any complex tables in design that don't use the first row as headers.

@kylepratuch , here's a nice to have for the ComplexTable component:
* Allow user to set vertical alignment on the table ("baseline" or "middle"). Currently assumes vertical alignment based on if the table is inside a table group
* I was going to suggest a row color dropdown (like simple table), but now that I look there's a lot of color options here... I still think it would be better than a hex value... Is there something like the color picker in October, where the color is shown but it isn't associated with a name?

dananjohnson:
All sounds good to me, @jen-castiron! And I agree about adding a color dropdown—that'd be really nice.

dananjohnson (approved):

PR #149, opened 2022-04-19
[F] Add investigation landing navigation

dananjohnson:


dananjohnson:
@jen-castiron give this a try if you don't mind! I've located the markup and styles in a new component, `StepNavigation`. `GuideNavigation` and `ChildNavigation` became thin wrappers around it that shape the link properties (guide and child navs get different prop shapes) and pass values for `columns` and `expandable`. I've tested both to make sure they work, but a second pair of eyes would be good. Thanks!

components/auth/AuthorizePage/index.js
```diff
@@ -48,39 +46,39 @@ export default function AuthorizePage({ typeHandle, children }) {
     router.push("/");
   }, [isAuthenticated]); // eslint-disable-line react-hooks/exhaustive-deps
 
-  // Show child content if authorized
```
jen-castiron, line 51: On a quick glance this code looks ok to me, but did you mean to comment out the educator page restriction?

dananjohnson: Oh, I just fiipped them around. So the conditional restricted message rendering logic comes first. I don't know if changes to this file should be committed. But I ran into a really weird issue with how `children` were being rendered from this component on the investigation landing page, and reversing the order of these `return` statements was the only way I could get it to work. 🤷🏻

dananjohnson: Oh I commented out the whole thing! 🙃 Sorry, misunderstood your comment. Yeah, that change can't be committed. I'll just remove it and we'll see if the issue resurfaces.

PR #150, opened 2022-04-20
[F] Add role buttons to sign-in modal

dananjohnson:


jen-castiron (changes requested):
👀 It looks like the translation for `investigation.coming_soon` was removed from en.json, and is probably why there's a conflict in es.json

jen-castiron (approved):

PR #156, opened 2022-04-21
[F] Add investigations to related content

jen-castiron:
- Adds investigations to related content
- Uses landingPage uri for uri
- No description field on investigations yet, will need to be added to the backend/API first

### DIFF
--- a/components/content-blocks/GridBlock/MainGrid.js
+++ b/components/content-blocks/GridBlock/MainGrid.js
@@ -45,13 +45,18 @@ const MainGrid = ({ items = [], limit, listTypeId, sectionHandle, pageId }) => {
                 mixedLink,
                 externalUrl,
                 uri,
+                landingPage,
               },
               i
             ) => (
               <Tile
                 key={id}
                 image={image?.[0]}
-                link={mixedLink || { url: externalUrl || uri }}
+                link={
+                  mixedLink || {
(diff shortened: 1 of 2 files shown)


dananjohnson (changes requested):
@jen-castiron I believe `description` has been added to the API now. I also think Kyle configured investigations to return `landingPage.uri` as its own `uri`. That might simplify the frontend a bit.

blnkt:
sorry I broke your stuffs with my force pushings, but Lucas thinks a rebase will getcha back on track A-ok.

dananjohnson:
@jen-castiron I've rebased your branch to fix the merge conflicts. See my earlier change request before merging this in. Thanks!

jen-castiron:
@dananjohnson Updated with description and uri! However, the uri is returning as `null` so I kept the landing page in as a fallback. @kylepratuch Should I expect the investigation's landing page to be returned as the top level uri? Or is it `null` expected?

dananjohnson (approved):
Giving my 👍🏻 on this. We'll see what Kyle says, but I think the fallback makes sense.

kylepratuch:
> @kylepratuch Should I expect the investigation's landing page to be returned as the top level uri? Or is `null` expected?

I had hoped you would get the landing page based on the Investigation configuration, assuming that relation exists. I'll double check that and follow up.

PR #169, opened 2022-05-16
[B] Hide "Read More" buttons from screen readers

### LINKED ISSUE #73: Ensure headings and labels are descriptive and unique
## Description
When headings and labels are used, they must describe the topic or purpose of the element or content that they pertain to. Headings and labels must be unique unless there is sufficient context to allow users to differentiate between duplicated headings or labels. When headings and labels are not descriptive, users with cognitive impairments may have trouble finding the correct section. Users with visual impairments may have difficulty skimming the material to determine the correct section to review. Being able to predict the correct section where content will appear will save users who read slowly time and allow them to read only relevant content.

### News Grid
- [ ] "Read More" buttons should be hidden from screen readers

43ec81f2f154e4016a39d216e404c5eb668f1148-1641299497

ellajtanttu:
Resolves #73

### DIFF
--- a/components/atomic/Tile/index.js
+++ b/components/atomic/Tile/index.js
@@ -74,7 +74,7 @@ const Tile = ({
       {footer && (
         <div className="footer">
           {footer.button && (
-            <button aria-hidden="true" className={`c-buttonish`}>{footer.button}</button>
+            <div className={`c-buttonish`}>{footer.button}</div>
           )}
           {footer.sticker && <div>{footer.sticker}</div>}
         </div>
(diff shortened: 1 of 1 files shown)


dananjohnson (changes requested):
@ellajtanttu It's a good solution and the right one to solve the problem described in the issue. That said, I'm seeing an additional issue that they didn't account for, which is that this is a `button` element nested inside of an `a` element. Nesting interactive elements is an a11y no-no, and in this case pointless, because the `button` doesn't do anything. If we leave it though, it is focusable even if hidden to screen readers. So can you just change it to a `div` (and remove `aria-hidden`)? Let me know if any of that doesn't make sense.

dananjohnson (approved):
Approved! Make sure to use **Squash and Merge** when you merge this in.

PR #170, opened 2022-05-16
[B] Update unnecessary h4 tags to divs in Event Page

### LINKED ISSUE #68: Avoid unnecessary use of heading elements
## Description
Heading elements should be utilized only when necessary and primarily to mark the beginning of related sections of content. Making an effective use of headings can provide an excellent benefit to users of assistive technology, which may provide the user with a way to navigate section-by-section through content. By applying headings incorrectly, or for passages of text which do not delineate a section, users of assistive technology may be stripped of this useful navigation technique.

### Event Page
- [ ] Event Page Pretitle should be styled as it is, but should be a `div` instead of an `h4`.

21017dc62613ef4e209076272b9e37a332dc01e6-1641292344

### Staff Profile Page
- [x] Staff Profile Pretitle should be styled as it is, but should be a `div` instead of an `h4`

cd129dea3eaf0c94f4d12b045595380af828611c-1641219067

## CIC Estimate
0 points

ellajtanttu:
Resolves #68

### DIFF
--- a/components/templates/EventPage/index.js
+++ b/components/templates/EventPage/index.js
@@ -73,7 +73,9 @@ export default function EventPage({
       <Container paddingSize="medium">
         <div>
           {eventType?.[0]?.title && (
-            <Pretitle>{eventType?.[0]?.title}</Pretitle>
+            <Pretitle className="t-heading-quaternary">
+              {eventType?.[0]?.title}
+            </Pretitle>
           )}
           <h1>{title}</h1>
           <Subtitle>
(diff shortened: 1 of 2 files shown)


dananjohnson (changes requested):
@ellajtanttu you solved the a11y issue with the HTML, but it looks like there was a style regression.

_Before:_
Screen Shot 2022-05-16 at 4 52 06 PM

_After:_
Screen Shot 2022-05-16 at 4 52 27 PM

There must have been some styling that the `h4` tag  inherited that got lost when it was switched to a `div`.

Also, they didn't note it, but I'm seeing the same `h4` tag on the news detail page (`components/templates/NewsPage/index.js`). Let's go ahead and fix it there as well.

PR #182, opened 2022-05-25
[B] Update grid and pagination links to be `ul`

ellajtanttu:


dananjohnson (changes requested):
Great work, @ellajtanttu! All the markup looks correct. I noticed one small styling issue, which is that the space between pagination numbers seems to have been lost.

_Before:_
Screen Shot 2022-05-25 at 1 18 17 PM

_After:_
Screen Shot 2022-05-25 at 1 17 03 PM

You can solve this with a `gap` on the `StyledPaginationList` component.

There's also a merge conflict in the Tile component. If you need any help fixing that, let me know!

dananjohnson (approved):

PR #186, opened 2022-06-14
[C] update investigation fragment to use DAM

alexgoff:


blnkt (approved):
I think this did the trick.

PR #220, opened 2022-08-12
[B] Fix sign up password field

jen-castiron:


components/form/AuthModalPasswordField/index.js
```diff
@@ -16,7 +17,7 @@ const PATTERN = [
   "^_`{}|~-]){1,}).{8,}$",
 ].join("");
 
-export default function PasswordField(props) {
+function PasswordField(props, ref) {
```
jen-castiron, line 20: The password field was returning the value "undefined", which broke the sign up form! React-hook-form needs the input ref to work properly, so this quick change will fix the issue. @kylepratuch

ericdrosas87 (approved):

PR #226, opened 2022-09-13
[B] Set max-width for mobile nav & subnav

### LINKED ISSUE #200: Mobile nav items should be reachable at all screen widths
## Expected Behavior
Mobile nav item button text should be readable and reachable at all screen widths.

## Actual Behavior
On screens narrower than ~420px nested nav items are cut off horizontally

Image

dananjohnson:
Resolves #200


blnkt (approved):
That does it.

PR #227, opened 2022-09-14
[F] Implement RSS feeds for various entry types

dananjohnson:
@blnkt, I've created a proof-of-concept for generating an RSS feed from a Craft entry type in Next.js. Following the tutorials I could find, I set up an async function inside the `getStaticProps` method of the homepage that gets some data on staff entries from Craft and generates RSS, Atom, and JSON files. Those files are then written to the public directory. It all seems to work pretty seamlessly!

I had a number of questions come to mind while building this that will need to be resolved before proceeding:

1. **What page does this live on?** Since we're hooking feed generation into `getStaticProps`, we have to figure out what page to put this on. For the POC I just used the homepage, but that's probably not right. What if we create a new route in the frontend, like `/feeds`, for this? We could show a list of all the feeds with links since you have to render something. This page could regenerate at a set interval, but an even cooler thing would be to bind regeneration to Craft webhooks. Or even better, we set up a directory of feeds pages, with a page per entry type, and the `/feeds/staff` page, for instance, regenerates whenever a staff profile is created or updated. Maybe not an MVP feature, but just want to get the idea out there.
A related option would be to create a new page entry subtype in Craft, called "Feed Page" or something. It could be the page where all feeds are generated, or you could perhaps select which entry type to generate feeds for from a preset list. The feed files would still probably need to be written to the `/public` directory, so I'm not sure how useful this would be to you, but you'd at least have control over where the feed index page(s) lived.
2. **What data do we show for an entry item?** Each entry in the feed can have the following data: title, url, description, content, authors, contributors, date updated, and image. Largely straightforward and simple to pull from Craft, but I'm not sure what to do for content and description. Typically content would be the full HTML of the entry, but that's not really something we can access via graphql, nor would it necessarily make sense. Would you render content blocks in an RSS feed? So there are a couple options: 1) decide what constitutes the `content` property for each entry type for which you want a feed. For staff profiles, for instance, it could be the contents of the bio field. Or, 2) we omit the `content` property altogether and just provide the `description` property. That, of course, opens up the question of what counts as the description. 🙂 Do we need a new field on some or all of the entry types for this?
3. And finally, simply, **what entry types will have feeds?**

Interested to get your input on these issues!


blnkt:
Excellent POC.  Looks like you found the package for the job and the rest was just sticking it all together.  Cool. 

1. I like where your head is at, and I think organizing routes like `feeds/<TYPE>` feels right to me, however, I don't I see a use value in a webpage dedicate to such.  Instead I wonder if this is a good use case for some [next.js API routes](https://nextjs.org/docs/api-routes/introduction), or even a [dynamic API route](https://nextjs.org/docs/api-routes/dynamic-api-routes) such that we ended up with a file like `/api/feeds/[entryType].js` that could spit out whichever batch of entries.  In this way, I'm also wondering out loud if for feeds it's appropriate for the resources to be static (as you say, a file that gets updated whenever there are new entries of that type available), or if it makes more sense for these listings to be provided on demand.  And honestly "I'm not totally sure one way or the other" but I guess I'm intuitively leaning toward "on demand."  I also think it would be useful it there was an "all" feed that aggregates all the entry types we serve up as feeds in date sorted order.
2. I'm trying to read some like "making your own RSS feed" best practices, and it seems like, in case, we should probably just omit `content` for now and include `title`, `url`, `description`, `date updated` (if available), and `image`.  This leaves these feeds more as like "previews" to grab the user and drive traffic to the site itself, which seems sufficient to me.  What populates these fields probably varies slightly for each entry type...
3. `news`, `staff-profiles`, `events`, `jobs`, and as mentioned above, I think an `all` that's all of those types combined into a single feed would be nice.  In the interest of your available time and our available funds, I'd like to just focus on `news` and `staff profiles` right now, with an eye for you implementing work in a generalized/abstracted enough way that I can somewhat easily dig in here later and expand to the other entry types I've in mind.  For those entry types map the following craft fields to the corresponding feed fields:
**News**
- url: the full url of the post
- date updated: date
- title: title
- image: hero
- description: teaser

**Staff Profiles**
- url: the full url of the post
- date updated: I guess the craft `post date` if available, otherwise 
- title: dash separated combo of name, title, and location i.e. `<name> – <title> – <location>`
- image: portrait
- description: bio

Ok.  Lengthy response, I know, but I hope gives you a clear path with concrete deliverables to work toward.  If any more discussion needed don't hesitate to reach out.  I'm on parental leave every Monday & Tuesday this month but will make a point to check slack and email and stuff at least once a day and happy to chat through on a call later in the week if helpful.

dananjohnson:
Super helpful response, @blnkt. Just what I need to move this feature forward, thanks! I LOVE the idea of a dynamic API route. I was hung up on the conventional Next.js way of thinking (e.g. routes for page) that I never even thought of this approach! Super cool. I'll take it for a spin and see how feels—stay tuned.

dananjohnson:
@blnkt I've moved feed generation to a dynamic API route and configured support for staff, news, and events feeds. You can try generating one by hitting the `/api/feeds/${entryType}` endpoint. I looked into adding support for job entries, too, but it wasn't clearcut how to map the entry data to the feed item properties. That said I've set this up to be straightforward to add new entry types: you just add the name to the array of supported types, create a graphql fragment for entry data, and add a method to map the data to feed item properties if needed.

You also mentioned an "all" feed, and because dynamic API route directories require an index route, it made sense to me to add it there. So if you hit `/api/feeds` it'll generate a single feed from all supported types.

Have a look and let me know if you would suggest any changes or want to see any other features. Thanks!

dananjohnson:
@blnkt one thing that occurred to me just as that meeting was ending is that we should have our Craft webhooks hit these API routes to regenerate the feeds! Maybe this is already what you had in mind?

blnkt:
I hadn't thought that far ahead, but you're totally right, that's an excellent way to keep the feeds up to date with most recently published entries.  I'll merge this PR in for now and you can look into to updating your ISR branch to include this functionality?

blnkt (approved):

PR #228, opened 2022-09-14
[EPO 6872] EDC Logger

ericdrosas87:


blnkt (changes requested):
I understand the eye for reusing an existing fetch method here, but I'll suggest that with what you've got going on there + what your new `setEdcLog` method is doing, it makes more sense to pull that `edc-log.js` file out of the `api` dir and stick it at the same level as `fetch.js` (`lib`) and just have it do the whole process independent of any changes to `fetch.js`.  That's my 2 cents.  My rationale being that `fetch.js` and `queryAPI` feel particularly concerned with fetching data from the Craft API, whilst your logging method is particularly concerned with pushing data to a non Craft API resource.

blnkt:
Oh also @ericdrosas87 I squashed and rebased this branch so it could cleanly sit on top of `develop` without merge commits.  So you'll need to `git reset --hard origin/EPO-6872` from your local before doing anything more to this branch.

ericdrosas87:
> I understand the eye for reusing an existing fetch method here, but I'll suggest that with what you've got going on there + what your new `setEdcLog` method is doing, it makes more sense to pull that `edc-log.js` file out of the `api` dir and stick it at the same level as `fetch.js` (`lib`) and just have it do the whole process independent of any changes to `fetch.js`. That's my 2 cents. My rationale being that `fetch.js` and `queryAPI` feel particularly concerned with fetching data from the Craft API, whilst your logging method is particularly concerned with pushing data to a non Craft API resource.

Works for me, I'll make that change this afternoon

blnkt:
@ericdrosas87 I'm still seeing the same `Error: undefined` messages in [this branch's build log](https://github.com/lsst-epo/rubin-obs-client/actions/runs/3116244348/jobs/5065060335).  I [reran a recent `develop` build](https://github.com/lsst-epo/rubin-obs-client/actions/runs/3109474819/jobs/5065011802) and saw no such errors.  I don't see the error messages on my local when building your branch, even when building against the dev env API.  Which is all to say, the `Error: undefined` message feels specific to this most recent work, and potentially also to deployments only.  I'm happy to troubleshoot this next week if the issue is feelin' too frontendy.  But I'd rather not merge this work in until it is building free of error messages (erroneous or not).

ericdrosas87:
> @ericdrosas87 I'm still seeing the same `Error: undefined` messages in [this branch's build log](https://github.com/lsst-epo/rubin-obs-client/actions/runs/3116244348/jobs/5065060335). I [reran a recent `develop` build](https://github.com/lsst-epo/rubin-obs-client/actions/runs/3109474819/jobs/5065011802) and saw no such errors. I don't see the error messages on my local when building your branch, even when building against the dev env API. Which is all to say, the `Error: undefined` message feels specific to this most recent work, and potentially also to deployments only. I'm happy to troubleshoot this next week if the issue is feelin' too frontendy. But I'd rather not merge this work in until it is building free of error messages (erroneous or not).

Yeah that's fine, I'll take a look at the Github Action build when I have some downtime

ericdrosas87:
Found the issue, the env var wasn't being set for the edc-log call. The error was being swallowed because it's actually a stack trace, so I moved it to its own `console.warn()` line for both the `edc-log.js` and `fetch.js` files. Also had to add the env var to the workflow files and `Dockerfile`.

PR #235, opened 2022-10-06
[F] Build Gallery pages from Canto API

dananjohnson:
@blnkt, I think this is at a good point for you to review. I haven't implemented the filter/search functionality yet. But otherwise the paginated gallery results and detail pages are both totally based on the Canto API now.

There are a couple categories of things to discuss with this first pass. The first is the new dynamic routes in Next; the second is with the asset data we get back from Canto, and how that lines up (or doesn't) with how this was previously built with Craft.

First, since the asset detail pages are derived exclusively from Canto and will have no record in Craft, I had to create a new dynamic route for them in Next. (Two actually: `gallery/[id].js` and `es/galeria/[id].js`.) For now I left the gallery index pages as Craft page entries. Those could be moved out as well, but I wasn't sure about that since it would remove the slideshow at the top and prevent any other Craft content from being added to the page. A downside of using both Canto and Craft is that I had to hardcode the gallery path from Craft into the the dynamic detail route path. So should the gallery index page ever move from `/gallery`, it'd no longer be a direct parent. Maybe that's not a big concern though? The gallery root page, which is used to generate the backlink and breadcrumb links on the detail pages, could always be adjusted in Craft.

OK, the Canto asset data:

There are a handful of metadata fields we had in Craft (and our frontend template) that don't line up with what we get back from Canto. Can you tell me how to handle these things?
 - `location` (no straighforward field in API response)
 - `metadataDate` (`metadata["Date modified"]` or …?)
 - `publisher` (probably `additional["Publisher"]`?)
 - `variants` (nothing exactly in API response)

On a related note, we had originally set up fields in Craft to supply default values for some of these. As I've built this now, we still query for and use those values as a fallback. Should we keep that, or use Canto exclusively?
```gql
... on galleryItemDefaults_GlobalSet {
  creditDefault
  metadataVersionDefault
  publisherDefault
  publisherIdDefault
  usageTermsDefault
}
```

Let's talk about media types lastly. I could only find images in the Canto album I tested with. Do you want to support any of the other Canto schemes as well ("video", "audio", "document", "presentation" or "other")? For images, the API provides just two direct image URLs: `directUrlPreview` and `directUrlOriginal`. I initially decided to use the full-resolution `directUrlOriginal` image and leverage Next's Image component to generate a set of optimized versions for the website. But that's leading to some broken images (see page 2 of results) when the original file is a TIFF or other non-browser-friendly format. The problem with the `directUrlPreview` file is that it's only 800px wide, definitely too small for the detail page as well as for larger thumbnails. There are other formats in the API response (see the `url` property), but all require an additional API request. Maybe that's what we have to do though?

I'll leave it there and await your feedback. In the meantime, I'll get started on the filter/search functionality.


dananjohnson:
Btw until we have the Craft part of this working, you'll need to use an env var (`CANTO_ACCESS_TOKEN`) to supply a valid access token.

blnkt:
Hardcoded gallery/galeria feels like a fine thing.  Unlike other interior pages, the gallery is such a pillar of the site I can't imagine ever burying it any deeper in the page tree, or changing its name to anything too wild.

For metadata fields, please rely exclusively on Canto response (i.e. don't use the defaults, Canto has its own sense of default we should be leveraging).  If a field is null, please don't show it.  For the 4 you called out:

- `location`: ditch it for now
- `metadataDate`: please use `metadata["Metadata Date"]`.  This one will be missing for a lot of files as we're still early in the process of updating the exif metadata on our images.  This image has probably the most full set of metadata of what we have up in the gallery currently: `http://localhost:3000/gallery/ai03ln71vt1bl84fu316jig871`. Which may help for your testing.
- `publisher`: yep, use `additional["Publisher"]`.  Is another one that'll be blank in many cases as we continue to catch up with our data entry.
- `variants`: ditch it for now

In addition to the above, please show: 

- Spatial Coordinate Frame: `additional["Spatial Coordinate Frame"]`
- Spatial Coordinate System Projection: `additional["Spatial Coordinate System Projection"]`
- Spatial Reference Dimension: `additional["Spatial Reference Dimension"]`
- Spatial Reference Value: `additional["Spatial Reference Value"]`
- Spatial Rotation: `additional["Spatial Reference Rotation"]`
- Spatial Scale: `additional["Spatial Scale"]`

These are mostly only used in scientific images, so will be null for a lot of images, but are important o show in the list of metadata where available.

For rendering the asset on the screen, I've 2 thoughts.  1) For the reason you described, I'd like to use the `directUrlPreview` for thumbnails for now (e.g. in the grid).  And use the `LowJPG` for the details page.  And `directUrlOriginal` for the download button on the details page.  2) After a little experimentation, it looks to me like the last param in the `directUrlPreview` url is actually generating whatever size as specified "on the fly", always a jpeg.  For instance, https://Rubin.canto.com/direct/image/ai03ln71vt1bl84fu316jig871/4LdGEkGJLCZEXwYaxSeGxWsHPmY/m800/800 and https://Rubin.canto.com/direct/image/ai03ln71vt1bl84fu316jig871/4LdGEkGJLCZEXwYaxSeGxWsHPmY/m1400/1400 both return images sized as expected.  Which is actually kinda neat. 
So, since our grid sizes are consistent (i.e. 1-up, 2-up, 3-up, 4-up sizes), perhaps use the `directUrlPreview`, but modified as needed in the grid.  And/or, could leverage this to use the `url`, `url2x`, `url3x` stuff we had going on with our image component previously?  I don't know thinking outloud a bit on this last point.

Finally, in terms of what kinds of assets we hope to include.  We will have images, videos, and 3D renderings.  Images work the way we expect.  I think videos let's skip for now and I need to explore how best to juggle between canto & youtube.  However, [there is one video in that album](http://localhost:3000/gallery/ka1d880url41v7cab4o0n6it53) if that's helpful for testing.  For the 3D renderings, they're zip files, which seem to fall under a kind canto calls `other`.  They do still use a thumbnail image as their `directUrlPreview` so that's probs another good reason to use that for the grid.  Here's one example of [a 3D file]( http://localhost:3000/gallery/eq81jr7mt544j45rggb9oci37a).  So yea, I think at this point if you can puzzle through having At least images and the 3D files showing up in the thumbnail grid properly, and their details pages working, that'd be great.  If easy to incorporate videos too even better.

dananjohnson:
Thanks for all this helpful feedback, @blnkt! I will take another pass at it early next week, and get search and filters wired up as well.

Good call on passing size values in the `directUrlPreview` URL! If we do that, do you still want to fetch the `LowJPG` version for the detail page?

blnkt:
¯\_(ツ)_/¯

yes?  I guess just to have accepted Canto's definition of "low"?

dananjohnson:
@blnkt I've done a second pass at what's rendered for the gallery assets. I was able to get support for other media types working by revising the dynamic route slightly from `/gallery/[id]` to `/gallery/[scheme]/[id]`. This was necessary because the scheme has to be included in the API request. But it meant that with just a few small rendering changes we have support for videos and 3D models ("other") in place. We could easily adjust things later if you decide to handle video display through YouTube.

For images, I found that I could generate a `srcset` for each image from the `directUrlPreview` property. Getting a different size by changing the last segment of that URL proved to be a bit more limited than we hoped—the options are 320, 500, 640, 800, or 2000px wide—but it seemed like a solid range to work with. I also looked into LowJPG but the returned image was 3000px wide; better to save an API call and just use the set from `directUrlPreview`, I thought, but I'm happy to change that if you feel otherwise.

The metadata properties basically are all showing up as expected, with the exception of `metadata["Date Created"]`. The returned format looks like `2021:05:27`, which JavaScript's `Date` object doesn't really know what to do with. (By contrast, `metadata["Metadata Date"]` returns a nice, readable timestamp.)

I think the last thing I have to do is revise the filter/search interface to work with our new setup. I'm gonna turn my attention to that now.

dananjohnson:
@blnkt filter/sort/search is in place too now. I think this is all set for a final review!

PR #246, opened 2022-12-15
Sibling Nav siblings are only siblings is they have the same parent

blnkt:
### Expected Behavior
On pages that have the "sibling nav" enabled, the "next" and "previous" links are disabled if you are on the first or last sibling page (respectively). 

### Actual Behavior
In some cases, Craft CMS considers pages to be siblings even if they do not have the same parent.

### Fix
Adds filter param to `getSiblings` graphQL query that only includes `prev` or `next` siblings in the response if their parent entry id matches the current page's parent id (i.e. the siblings and the current page all have the same parent).  Additionally, if a page has no parent it is assumed it has no siblings and therefore query is not made and `addSiblings` returns `{}` instead of querying the CMS.

### Testing
Please test all page types:
- A Page with the Sibling Nav enabled.  For instance the previous button should be grayed out on http://localhost:3000/education/educators/investigations/expanding-universe/teacher-guide-exu/introduction and the next button should be grayed out on http://localhost:3000/education/educators/investigations/expanding-universe/teacher-guide-exu/diversity-equity-and-inclusion.
- A page without the sibling nav enabled. For instance http://localhost:3000/about should render normally without the sibling nav
- Homepage should render normally without the sibling nav
- Channel based dynamic landing page like news, events, or rubin voices should render normally without the sibling nav

### DIFF
--- a/lib/api/sibling-nav.js
+++ b/lib/api/sibling-nav.js
@@ -4,22 +4,23 @@ import { queryAPI } from "@/lib/fetch";
 export async function addSiblings(entryData, site = "default") {
   if (!entryData) return null;
 
-  const { uri, level } = entryData;
-
-  const siblings = await getSiblings(uri, level, site);
-
+  const { uri, level, parent } = entryData;
+  const parentId = parent?.id;
+  const siblings = parentId
+    ? await getSiblings(parentId, uri, level, site)
+    : null;
   return Object.assign(entryData, { siblings });
 }
 
-export async function getSiblings(uri, level = 1, site = "default") {
+export async function getSiblings(parentId, uri, level = 1, site = "default") {
   const query = gql`
       {
         siblings: entry (uri: "${uri}", site: "${site}") {
-            prev(section: "pages", site: "${site}", level: ${level}) {
+            prev(descendantOf: ${parentId}, section: "pages", site: "${site}", level: ${level}) {
                 uri
                 title
             }
-            next(section: "pages", site: "${site}", level: ${level}) {
+            next(descendantOf: ${parentId}, section: "pages", site: "${site}", level: ${level}) {
                 uri
                 title
             }


lib/api/sibling-nav.js
```diff
@@ -4,22 +4,23 @@ import { queryAPI } from "@/lib/fetch";
 export async function addSiblings(entryData, site = "default") {
   if (!entryData) return null;
 
-  const { uri, level } = entryData;
-
-  const siblings = await getSiblings(uri, level, site);
-
+  const { uri, level, parent } = entryData;
+  const parentId = parent?.id;
```
blnkt, line 8: conditionally pull out the parent ID (conventionally we use URI as entry identifier but we have to use page ID here because that's what the graphQL filter accepts

lib/api/sibling-nav.js
```diff
@@ -4,22 +4,23 @@ import { queryAPI } from "@/lib/fetch";
...
+  const siblings = parentId
+    ? await getSiblings(parentId, uri, level, site)
+    : {};
```
blnkt, line 11: If there isn't a parentId then don't attempt to grab the siblings

lib/api/sibling-nav.js
```diff
@@ -4,22 +4,23 @@ import { queryAPI } from "@/lib/fetch";
...
   const query = gql`
       {
         siblings: entry (uri: "${uri}", site: "${site}") {
-            prev(section: "pages", site: "${site}", level: ${level}) {
+            prev(descendantOf: ${parentId}, section: "pages", site: "${site}", level: ${level}) {
                 uri
                 title
             }
-            next(section: "pages", site: "${site}", level: ${level}) {
+            next(descendantOf: ${parentId}, section: "pages", site: "${site}", level: ${level}) {
```
blnkt, line 23: In a perfect world the underlying Craft shenanigans ought to be constraining this `prev` and `next` already, but appears they are not.  The edge case seems to be weird stuff like entries who share a parent but it's not a "direct" parent, or entries used to be siblings but were reshuffled in the pagetree, and some other stuff like that.  Adding this filter seems to affirm strongly enough to only return "direct" siblings 🤷

lib/api/sibling-nav.js
alexgoff: I don't see any issue but I think `null` could be a better alternative than `{}`

The `SiblingNavigation` component expects a `siblings` object with `next, prev` properties OR it returns `null` if `siblings` is a falsey value. It probably doesn't really matter since afterwards it checks for the existence of `next,prev` anyway, and the parent needs to have children for the sibling nav to show anyway.

alexgoff (changes requested):
`null` is probably a better value for the `siblings` property but I don't think it makes a significant difference since there are other catches that will prevent unwanted behavior.

blnkt:
Good call.  Changing now.

PR #250, opened 2022-12-22
[C] replace icons with epo-react-lib

alexgoff:
This solution works but I'm not completely happy with it, this discussion should probably move over to epo-react-lib though. 

Ideally what I would like is that `icons` is a sub-module of `epo-lib` à la 

```
import { Account } from '@rubin-epo/epo-react-lib/icons'

export default Account
```

This opens up a can of worms with Vite library multi-entrypoints, typings for multiple entries, and then finally that the multi-entry setup requires using the `.mjs` extension which forces ES6 module resolution which requires file extensions on imports. Definitely something I am open to exploring since I don't like having to double destructure, nor do I want to have to manually export each icon. It would also open up other sub-modules like `atomic` `layout` etc.


blnkt:
[My commit](https://github.com/lsst-epo/rubin-obs-client/pull/250/commits/02c34029c95482b6f008cda5c907006c84bfe43c) dumps the icons dir completely, refactoring the filterbar and expand/collapse icon as needed.

components/page/FilterBar/index.js
```diff
@@ -143,7 +142,7 @@ const FilterBar = ({ filterType, setLiveRegionMessage }) => {
         </div>
         <FilterSearch onSubmit={handleSubmit}>
           <button type="submit">
-            <SearchIcon />
```
alexgoff, line 146: good standardization

alexgoff (commented):
no errors building or using the site locally

PR #251, opened 2022-12-29
EPO-7377

alexgoff:
Updated `epo-react-lib` to latest 1.0.12, pulled in the `Button`, `ExpandToggle`, and `Video` components. These components were barrel exported in `@/atomic/` so I imported them there and removed their individual component files along with some others.


components/atomic/index.js
```diff
@@ -1,12 +1,15 @@
...
-export { default as Button } from "./Button";
 export { default as SSOButton } from "./Button/patterns/SSOButton";
 export { default as Buttonish } from "./Buttonish";
-export { default as ExpandToggle } from "./ExpandToggle";
 export { default as Tile } from "./Tile";
 export { default as InvestigationTile } from "./Tile/patterns/InvestigationTile";
 export { default as SlideTile } from "./Tile/patterns/SlideTile";
 export { default as Share } from "./Share";
+
+export {
```
alexgoff, line 8: replaced exports in here with epo-react-lib versions and updated all references to this file so that the individual component re-exports can be removed.

components/content-blocks/Video/index.js
```diff
@@ -1,8 +1,12 @@
 import PropTypes from "prop-types";
 import Container from "@/layout/Container";
 import { Video, Figure } from "@/components/atomic";
+import { useIsMounted } from "@/hooks";
 
 export default function VideoBlock({ caption, url }) {
+  const isMounted = useIsMounted();
+  if (!isMounted) return null;
+
```
blnkt, line 9: Was able to quickly track down the hydration error to this video component.  Some conflicting reports as to whether or not a third party iframe _should_ throw a hydration error, but felt like not too much harm in adding this quick and easy check, using an existing hook.  Hook returns true when a component is "mounted" aka the first time `useEffect` fires.  And, as it happens, that only happens browserside, so, for us, it doubles as a cheap check of whether or not we're in the browser.  And if the Video/iframe only loads in browser then React sees no diff in DOM and VDom.

blnkt (approved):

PR #266, opened 2023-01-23
[C] update phase IV components

alexgoff:
For JIRA: "EPO-7529 #IN-REVIEW #comment update phase IV components"


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "rubin-client",
-  "version": "0.1.0",
+  "version": "1.2.5",
```
alexgoff, line 3: there's a version number for published releases so it seems like it would make sense to put it in the package.json too, right?

package.json
```diff
@@ -57,7 +57,6 @@
     "add": "^2.0.6",
     "classnames": "^2.3.1",
     "feed": "^4.2.2",
-    "flickity": "^3.0.0",
```
alexgoff, line 60: flickity is a dependency of the carousel which is now in epo-react-lib

blnkt (approved):

PR #269, opened 2023-01-25
[C] update phase V components

alexgoff:
For JIRA: "EPO-7563 #IN-REVIEW #comment update phase V components"


components/atomic/Tile/patterns/InvestigationTile/styles.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable */
 import styled from "styled-components";
-import BaseMixedLink from "@/atomic/MixedLink";
+import { MixedLink as BaseMixedLink } from "@rubin-epo/epo-react-lib";
```
alexgoff, line 3: I'd like to point this back to atomic/index.js like the others but for some reason aliasing that export as `BaseMixedLink` produced `undefined` but aliasing the export from `@rubin-epo/epo-react-lib` worked

components/atomic/Share/index.js
```diff
@@ -1,10 +1,12 @@
...
-import CopyUrlButton from "./CopyUrlButton";
-import FacebookButton from "./FacebookButton";
-import TwitterButton from "./TwitterButton";
-import EmailButton from "./EmailButton";
+import {
+  CopyUrlButton,
+  FacebookButton,
+  TwitterButton,
+  EmailButton,
+} from "@rubin-epo/epo-react-lib";
```
alexgoff, line 9: this entire `Share` component could not be extracted because it relies on `useGlobalData`, so the subcomponents were instead.

blnkt (approved):
Looks good.  I do get an error from the FacebookButton when I test on my local, but not when I try it at the preview url, so I'm assuming it's a localhost kinda prob...although I felt like it used to work even on my local.  Anyhow, something to keep an eye but not enough of a thing to prevent us from merging this work in.

PR #273, opened 2023-02-13
[C] replace SimpleTable and ComplexTable

alexgoff:
Replaced the `SimpleTable` and `ComplexTable` components with `epo-react-lib` equivalents. I noticed that a new prop had been added to determine if a table is site specific so instead of wholesale replacing the components they just wrap the library components with this new logical choice


blnkt (approved):
👍 SO clean now that all the table cruff lives inside the lib

PR #291, opened 2023-04-26
[F] WidgetGrid

alexgoff:
For JIRA: "EPO-7926 #IN-REVIEW #comment add WidgetGrid"

### DIFF
--- a/components/layout/WidgetGrid/WidgetGrid.stories.jsx
+++ b/components/layout/WidgetGrid/WidgetGrid.stories.jsx
@@ -1,0 +1,29 @@
+import styled from "styled-components";
+import WidgetGrid from ".";
+
+const meta = {
+  component: WidgetGrid,
+};
+export default meta;
+
+const MockWidget = styled.div`
+  border-radius: 6px;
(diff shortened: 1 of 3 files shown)


components/layout/WidgetGrid/WidgetGrid.stories.jsx
```diff
@@ -0,0 +1,29 @@
+import styled from "styled-components";
+import WidgetGrid from ".";
+
+const meta = {
+  component: WidgetGrid,
+};
+export default meta;
+
+const MockWidget = styled.div`
```
alexgoff, line 9: this example shows how to set up responsive grid columns, if a widget is configured with a span greater than the current number of columns, then CSS `min()` will compare the number of columns from the prop and the number of columns defined by the grid and use the lesser value.

blnkt (approved):
Nice.  Note to selves, this masonry-esque layout is very very reminiscent of what our image gallery layout used to be like (before we linked to the external Canto gallery).  In the future it may be worth abstracting or adapting this layout component to that use case as well.

PR #299, opened 2023-05-04
[F] Charting lib

alexgoff:
- Added charting tools to support Summit Status Dashboard, no new dependencies pulled in for charting. Dependency change comes from adding Storybook a11y addon.
- The ChartFigure component is the only component with a story and contains two examples: one with bar and line graph data, and the other with bar graph data but overflowing the parent container. The overflow occurs automatically and is based on the specified height of the chart.
- There is no singular "Chart" component where a config and some data are passed resulting in a chart; rather these are components intended to be assembled as needed into containers for charts. They may be mixed, matched, and enhanced with bespoke components as-needed, there is no prescribed "Bar Graph" or "Line Graph".

### DIFF
--- a/components/charts/BarData/index.js
+++ b/components/charts/BarData/index.js
@@ -1,0 +1,48 @@
+import PropTypes from "prop-types";
+import Bar from "../Bar";
+
+const BarData = ({
+  data = [],
+  xDomain,
+  yDomain,
+  xScale,
+  yScale,
+  ticks,
+  labelledById,
+  showTooltips,
+  tooltipFormatter,
+}) => {
+  const interval = (xDomain[1] - xDomain[0]) / (ticks - 1);
+
+  return (
+    <g role="list" aria-labelledby={labelledById}>
+      {data.map((value, i) => {
+        const height = yScale(yDomain[1] - value);
+        const x = xScale(xDomain[0] + interval * i);
+        const y = yScale(yDomain[0]);
+        return (
+          <Bar
+            {...{ height, x, y, value, showTooltips, tooltipFormatter }}
+            key={`${value}-${i}`}
+          />
+        );
+      })}
+    </g>
+  );
+};
+
+BarData.displayName = "Charts.BarData";
+
+BarData.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.number).isRequired,
+  xDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
+  yDomain: PropTypes.arrayOf(PropTypes.number).isRequired,
+  xScale: PropTypes.func.isRequired,
+  yScale: PropTypes.func.isRequired,
+  ticks: PropTypes.number.isRequired,
+  showTooltips: PropTypes.boolean,
+  tooltipFormatter: PropTypes.func,
+  labelledById: PropTypes.string,
+};
+
+export default BarData;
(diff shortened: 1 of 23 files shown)


components/charts/BarData/index.js
```diff
@@ -0,0 +1,48 @@
+import PropTypes from "prop-types";
+import Bar from "../Bar";
+
+const BarData = ({
```
alexgoff, line 4: map of Bars

components/charts/ChartBase/index.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from "prop-types";
+
+const ChartBase = ({ children, className, width = 900, height = 300 }) => {
+  return (
+    <svg
+      preserveAspectRatio="xMidYMid meet"
+      viewBox={`0 0 ${width} ${height}`}
+      className={className}
+      style={{ minHeight: height, height: "100%" }}
```
alexgoff, line 9: this combo of height and minheight overflow the chart if it's wider than it's parent, the viewBox maintains aspect ratio

components/charts/ChartFigure/ChartFigure.stories.js
```diff
@@ -0,0 +1,197 @@
...
+  return (
+    <ChartFigure {...args} legend={<ChartLegend legends={legends} />}>
+      <ChartBase {...{ width, height }}>
+        <XAxis
+          {...{
+            xDomain,
+            xScale,
+            y: height - padding,
+            ticks: 24,
+            labelFormatter: (value) => {
```
alexgoff, line 48: label formatters are a powerful way to adjust labels either using unit formatting or dynamically (see the "NOW" label below)

components/charts/ChartFigure/index.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from "prop-types";
+import * as Styled from "./styles";
+
+const ChartFigure = ({ caption, children, legend }) => {
```
alexgoff, line 4: the ChartFigure wraps charts into a semantic figure that describes the chart with a caption

components/charts/LineData/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+  const points = data.reduce((prev, curr, i) => {
+    const nextPoint = `${prev} ${xScale(xDomain[0] + interval * i)},${yScale(
+      curr
+    )}`;
+
+    return nextPoint;
+  }, "");
+
+  return <Styled.Line points={points} aria-labelledby={labelledById} />;
```
alexgoff, line 22: how can the line data be made accessible? It does not have the "list" format of bar graphs where individual datapoints can be read out. Points could be assigned to the line and be invisible by default but included in the accessibility tree.

components/charts/Tooltip/index.js
```diff
@@ -0,0 +1,57 @@
...
+}) => {
+  const label = tooltipFormatter ? tooltipFormatter(value) : value;
+  const width = label.length * 8;
+  const arrowWidth = 6;
+  const xCenter = x + width / 2;
+
+  return (
+    <Styled.Tooltip
+      transform={`translate(-${width / 2} -${height * 1.5})`}
+      opacity={isVisible ? 1 : 0}
```
alexgoff, line 21: use opacity one the tooltip so that it is always in the accessibility tree and serves as the label for the bar data.

blnkt (changes requested):
Feature looks good, but please address linting warnings/errors, and then we'll rebase it in. 
Screen Shot 2023-05-04 at 1 30 42 PM

PR #307, opened 2023-05-15
[F] weather icons

alexgoff:
Add weather icons in preparation for the summit status dashboard. A new icon composer has been added that extends the `epo-react-lib` component with these icons, use `UniqueIconComposer` to access the complete set of icons. Test in Storybook by observing the `UniqueIconComposer` story and checking the display of the new icons (weather, sun, moon)


components/svg/UniqueIconComposer.jsx
```diff
@@ -0,0 +1,23 @@
+import PropTypes from "prop-types";
+import { IconComposer } from "@rubin-epo/epo-react-lib";
+import UniqueIcons from "./unique";
+
+/** adds the unique icons for rubin-obs-client to the icons
+ *  already defined in @rubin-epo/epo-react-lib
+ */
+const UniqueIconComposer = (props) => (
```
alexgoff, line 8: simple reimplementation of the IconComposer that extends it with new icons and passes on other props.

components/global/Header/index.js
```diff
@@ -3,8 +3,8 @@ import PropTypes from "prop-types";
 import Link from "next/link";
 import useResizeObserver from "use-resize-observer";
 import { useTranslation } from "react-i18next";
-import Logo from "@/svg/unique/Logo";
-import LogoFullSize from "@/svg/unique/LogoFullSize";
+import Logo from "@/components/svg/unique/site/Logo";
+import LogoFullSize from "@/components/svg/unique/site/LogoFullSize";
```
alexgoff, line 7: a little reorganization to separate out the types of unique icons

blnkt (changes requested):
Builds fare ailing because eslint wants all .jsx files to have PropTypes defined.  You have 
```
className,
size = 24,
fill = "currentColor"
```
for all of these new weather icons.  It's not clear to me if you actually do even want the `fill` and `size` to come in as props in this case, but assuming one does, I suggest reimplementing a lil helper "shape" that probably got refactored out when the icons all migrated into the UI lib, and also there became typescript.  The shape `svg.js` file looked like this:
```
import PropTypes from "prop-types";

export const svgInternalShape = {
  className: PropTypes.string,
  size: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
  stroke: PropTypes.string,
  fill: PropTypes.string,
  svgProps: PropTypes.object,
};

const svgShape = PropTypes.shape({
  ...svgInternalShape,
});

export default svgShape;
```
And then, I can't quite remember but I think you can then import it where needed and either do something like:|
```
Thing.propTypes = {
  ...svgInternalShape,
};
```
or
```
Thing.propTypes = svgShape
```
something like that...?

blnkt (approved):

PR #309, opened 2023-05-22
[F] Unit localization toolbar

alexgoff:
Toolbar for picking local temperature and windspeed units. Controls state of the SummitStatus block which is put into a provider. Default selections are determined by the user's current locale.

Design
image

Client
image

### DIFF
--- a/components/content-blocks/SummitStatus/index.js
+++ b/components/content-blocks/SummitStatus/index.js
@@ -1,17 +1,41 @@
 import PropTypes from "prop-types";
+import { useState } from "react";
 import { Container } from "@rubin-epo/epo-react-lib";
+import { useTranslation } from "react-i18next";
+import NestedContext from "@/contexts/Nested";
+import UnitLocalization from "@/components/layout/UnitLocalization";
 import WidgetGrid from "@/components/layout/WidgetGrid";
 import WidgetPreview from "@/components/layout/WidgetPreview";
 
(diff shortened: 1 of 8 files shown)


components/content-blocks/SummitStatus/index.js
```diff
@@ -1,17 +1,41 @@
...
-    </WidgetGrid>
-  </Container>
-);
+const SummitStatus = ({ summitStatusLayout, widgetPreviews = [] }) => {
+  /** this logic should be changed to useRouter after i18n refactor */
+  const {
+    i18n: { language = "en" },
+  } = useTranslation();
+  const [windspeedUnit, setWindspeedUnit] = useState(
+    language === "en" ? "kn" : "m/s"
```
alexgoff, line 16: setting the English default as knots since imo it is the standard imperial unit for windspeed, but maybe mph makes more sense since our users probably aren't sailors or pilots.

components/shapes/svg.js
```diff
@@ -1,15 +1,11 @@
...
   fill: PropTypes.string,
   svgProps: PropTypes.object,
 };
 
-const svgShape = PropTypes.shape({
-  ...svgInternalShape,
-});
-
-export default svgShape;
+export default svgInternalShape;
```
alexgoff, line 11: simplified this, exporting as a shape caused issues in the icon composer

PR #311, opened 2023-05-24
[EPO-8053] Early Access Investigation flag

blnkt:
This PR includes updates to flag so called "Early Access" investigations with a lil orange flag.  This "status" supplants a sort of status that was previously implemented that just indicated whether or not an investigation was "active" (live) or "inactive" (coming soon).  Those existing statuses and their functionality and related styling are retained/unchanged.  The investigations Entries in the dev environment have already been status-ified, so you should be able to check the below without making any changes in the CMS (if reviewing the preview branch.  If you wanna review locally you will need `rubin-obs-api` to be on [EPO-8053](https://github.com/lsst-epo/rubin-obs-api/pull/153) and go into the CMS and update so/all investigations' statuses.  This PR also includes some update to prettier and eslint which hopefully get both of them working more effectively and predictably.

Check for:
- [InvestigationTile](https://epo8053-dot-skyviewer.appspot.com/education/investigations-2) and [InvestigationHero](https://epo8053-dot-skyviewer.appspot.com/education/educators/investigations/expanding-universe) visually match Jose's designs.  N.B. I rounded the corner of the flag when on Tile so better match the Tile's corner.  This isn't in the original desig, but was okayed by Jose.
- Existing "Active" and "Inactive" status styles are still correctly applied to [InvestigationTiles](http://localhost:3000/education/educators/investigations)
- Styles appear consistent across Safari, Firefox, and Chrome (and Edge?).
- Prettier & eslint scripts in package.json all still run as expected (may need to run `yarn` to get/update packages)

Screen Shot 2023-05-25 at 10 07 13 AM
Screen Shot 2023-05-25 at 10 07 04 AM


.eslintignore
```diff
@@ -1,3 +1,4 @@
 .next/*
 node_modules
 out/*
+storybook-static/*
```
blnkt, line 4: ignored

.prettierignore
```diff
@@ -0,0 +1,40 @@
...
+.env.production.local
+
+# vercel
+.vercel
+
+# ide
+.vscode
+
+# storybook
+/storybook-static/*
```
blnkt, line 40: prettier docs said prettier ignore should be based on gitignore so.....I made it the same?

.prettierignore
blnkt, line 1: prettier docs said prettier ignore should be based on gitignore so.....I made it the same?

.prettierrc.json
blnkt, line 1: I didn't actually end up seeing any need to configure beyond the default behavior, but I'm sure we'll need to refine its rule set or what have you sometime down the line, and here's where we'll do that.

components/atomic/Flag/index.js
```diff
@@ -0,0 +1,47 @@
...
+          <g transform="matrix(1, 0, 0, 1, 0, 0)" filter="url(#Subtraction_6)">
+            <path
+              id="Subtraction_6-2"
+              data-name="Subtraction 6"
+              d="M100,120h0L50,96.587,0,120V96.449H100V120Z"
+              transform="translate(0 -96.45)"
+              fill={color}
+            />
+          </g>
+        </Styled.FlagTail>
```
blnkt, line 33: This is a styled svg.  I thought about putting it alongside the other svg icons, but felt so bespoke to this component (which in turn didn't feel like it should be alongside the svgs because it's slightly more so than that) that I just inlined it here.

components/atomic/Flag/patterns/EarlyAccess/styles.js
```diff
@@ -0,0 +1,16 @@
+import styled from "styled-components";
+import { fluidScale, tokens } from "@/styles/globalStyles";
+import Flag from "@/atomic/Flag";
+
+export const EarlyAccess = styled(Flag)`
```
blnkt, line 5: `Flag` is in charge of its shape, `EarlyAccess` is in charge of how stuff looks and lays out inside that shape.

components/atomic/Tile/patterns/InvestigationTile/styles.js
```diff
@@ -63,52 +127,34 @@ export const Title = styled.div`
...
+  top: var(--Tile-image-margin-top);
+  right: 16px;
+  left: auto;
 
-  &[aria-disabled="true"] {
-    --Tile-bg-color: #6a6e6e;
-    --Tile-image-filter: grayscale(100%);
+  @media (max-width: ${BREAK_DESKTOP}) {
+    right: 0;
+    border-top-right-radius: 16px;
```
blnkt, line 138: Gave it a lil border radius (not in the original designs) so it tucked into the corner of the tile a bit more seamlessly.  Was okayed by Jose.

components/atomic/Tile/patterns/InvestigationTile/styles.js
```diff
@@ -63,52 +127,34 @@ export const Title = styled.div`
...
+    top: 0;
+    right: auto;
+    left: 100px;
+  }
 
+  ${OffsetWrapper} {
     @media (max-width: ${BREAK_PHABLET}) {
-      --Tile-bg-color: transparent;
-      --Tile-color: #6a6e6e;
+      margin-left: -200%;
```
blnkt, line 149: I think this is a legit "css trick" kinda thing, but I didn't really test across browsers.  My "offset an element so that it can be positioned along its center axis" skills are a bit rusty.  Better more bulletproof way to accomplish this?

components/layout/InvestigationHero/styles.js
```diff
@@ -1,27 +1,38 @@
...
+);
+
+const BUTTON_WRAPPER_PADDING = fluidScale(
+  "60px",
+  "15px",
+  tokens.BREAK_TABLET,
   tokens.BREAK_MOBILE
 );
 
+const DURATION_WIDTH = "197px";
```
blnkt, line 21: Variable-izing so I could reliably use it to position the label

components/layout/InvestigationHero/styles.js
```diff
@@ -56,21 +67,73 @@ export const Text = styled.div`
...
+  top: 0;
+  right: ${DURATION_WIDTH};
+
+  @media (max-width: ${tokens.BREAK_TABLET}) {
+    right: auto;
+    left: 0;
+  }
+
+  ${OffsetWrapper} {
+    margin-right: -100%;
```
blnkt, line 114: I think this is a legit "css trick" kinda thing, but I didn't really test across browsers. My "offset an element so that it can be positioned along its center axis" skills are a bit rusty. Better more bulletproof way to accomplish this?

components/svg/UniqueIconComposer.jsx
```diff
@@ -1,5 +1,4 @@
 import { IconComposer } from "@rubin-epo/epo-react-lib";
-import svgShape from "@/shapes/svg";
```
blnkt, line 2: I think the svgShape only applies to svgs, not the composer.  The Proptypes for the composer would be more freeform and/or would be like the icon component string, not the svg porps?

package.json
```diff
@@ -15,11 +15,11 @@
     "start": "next start -p 8080",
     "bs": "next build && next start -p 3000",
     "lint": "npm-run-all lint:*",
-    "lint:js": "eslint -c .eslintrc.js \"**/*.{js,jsx}\"",
+    "lint:js": "prettier --check \"**/*.{js,jsx}\" & eslint -c .eslintrc.js \"**/*.{js,jsx}\"",
```
blnkt, line 18: Based on suggestions from eslint and prettier docs, we are now running prettier on its own instead as an eslint rule.  Conceptually eslint should now be solely linting for opinons about haw javascript is used, while prettier should be solely linting for Code Style (formatting in the file).

package.json
```diff
@@ -113,7 +113,6 @@
     "eslint-plugin-jest": "^27.1.3",
     "eslint-plugin-jsx-a11y": "^6.6.0",
     "eslint-plugin-n": "^15.2.4",
-    "eslint-plugin-prettier": "^4.2.1",
```
blnkt, line 116: Don't need it as a plugin if running on its own

styled.stylelintrc.js
```diff
@@ -75,6 +75,6 @@ module.exports = {
     "no-empty-source": null,
     indentation: null,
     "max-empty-lines": 3,
-    "no-eol-whitespace": true,
+    "no-eol-whitespace": null,
```
blnkt, line 78: was throwing too many false positives and seemed to be clashing hard with prettier config, so disabling this rule for now.

components/layout/InvestigationHero/styles.js
alexgoff: I think using 

```css
    translateX(-100%);
```

would be the cleanest way to accomplish this

alexgoff (approved):
Everything looks correct across a range of sizes in Chrome, Safari, and Firefox so I think this is good! I left one note on your comment about using `transform` instead of `margin-left` to position the element

PR #320, opened 2023-05-31
[F] add TemperatureCurrent widget

alexgoff:
Added TemperatureCurrent widget, shapes for the units (reverted back to "fahrenheit" and "celsius" since this is what JavaScript's `toLocaleString` expects as valid units)

### DIFF
--- a/components/content-blocks/SummitStatus/index.js
+++ b/components/content-blocks/SummitStatus/index.js
@@ -7,6 +7,7 @@ import UnitLocalization from "@/components/layout/UnitLocalization";
 import WidgetGrid from "@/components/layout/WidgetGrid";
 import WidgetPreview from "@/components/layout/WidgetPreview";
 import TemperatureCurrent from "@/components/widgets/TemperatureCurrent";
+import PrecipitationCurrent from "@/components/widgets/PrecipitationCurrent";
 
 const SummitStatus = ({ summitStatusLayout, widgetPreviews = [] }) => {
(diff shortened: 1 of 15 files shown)


components/content-blocks/SummitStatus/index.js
```diff
@@ -15,7 +16,9 @@ const SummitStatus = ({ summitStatusLayout, widgetPreviews = [] }) => {
   const [windspeedUnit, setWindspeedUnit] = useState(
     language === "en" ? "NM" : "m"
   );
-  const [tempUnit, setTempUnit] = useState(language === "en" ? "F" : "C");
+  const [tempUnit, setTempUnit] = useState(
+    language === "en" ? "fahrenheit" : "celsius"
+  );
```
alexgoff, line 21: JavaScript's number format library uses the long strings and convert.js can also use long strings so I reverted to those

components/widgets/TemperatureCurrent/index.js
```diff
@@ -0,0 +1,49 @@
...
+      .formatToParts(value)
+      .map(({ type, value }) => {
+        if (type === "unit") {
+          return value.replace(/[a-z]/gi, "");
+        }
+
+        return value;
+      })
+      .join("");
+  };
```
alexgoff, line 27: add a formatter on top of the JavaScript `Intl` lib to remove the "F" or "C" but preserve the degrees symbol and any other formatting applied

blnkt (approved):

PR #321, opened 2023-05-31
[F] add PrecipitationCurrent

alexgoff:
add PrecipitationCurrent widget and extract the background panel all widgets using into an atomic component

### DIFF
--- a/components/widgets/PrecipitationCurrent/index.js
+++ b/components/widgets/PrecipitationCurrent/index.js
@@ -1,0 +1,36 @@
+import PropTypes from "prop-types";
+import { useTranslation } from "react-i18next";
+import WidgetBackground from "@/components/atomic/WidgetBackground";
+import * as Styled from "./styles";
+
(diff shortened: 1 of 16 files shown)


components/widgets/PrecipitationCurrent/index.js
```diff
@@ -0,0 +1,50 @@
+import PropTypes from "prop-types";
+import { useTranslation } from "react-i18next";
+import WidgetBackground from "@/components/atomic/WidgetBackground";
+import * as Styled from "./styles";
+
+const PrecipitationCurrent = ({ precipitation = 0, humidity = 0 }) => {
+  const { t, i18n } = useTranslation();
+  const { language = "en" } = i18n;
+
+  const styledFormatter = (value) => {
```
alexgoff, line 10: make use of this custom formatter to add styling to the percent sign in keeping with the designs, while also preserving local percentage display

PR #329, opened 2023-06-12
[EPO-8165] Loading Spinner implementations

blnkt:
Loader graphic implemented across "dynamic" `dataList` and `gridBlock` components.  Loader should appear on page load, then be removed after successful data fetch.  Please review on pages with dynamic dataLists:

- /news
- /explore/voices/all-rubin-voices
- /calendar
- /gallery/slideshows
- /education/glossary
- /about/work-with-us/jobs (n.b. this page is currently disabled but there are jobs entries, so can be reviewed on local if page is reenabled)
- /search?search=space&page=1

and also on pages that have dynamic contentBlocks:
 - a news post (the "related" section at the bottom)
 - an event (the "related" section at the bottom)
 - a Rubin Voices page with a "related" block
 - a normal page with a "related" block
 - homepage (news block at the bottom)

### DIFF
--- a/components/atomic/Loader/index.js
+++ b/components/atomic/Loader/index.js
@@ -1,0 +1,9 @@
+import * as Styled from "./styles";
+
+function Loader(props) {
+  return <Styled.Loader {...props} />;
+}
+
+Loader.displayName = "Atomic.Loader";
+
+export default Loader;
--- a/components/atomic/Loader/styles.js
+++ b/components/atomic/Loader/styles.js
@@ -1,0 +1,8 @@
+import styled from "styled-components";
+import { CircularLoader } from "@rubin-epo/epo-react-lib";
+
+export const Loader = styled(CircularLoader)`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`;
--- a/components/content-blocks/GridBlock/CarouselGrid/index.js
+++ b/components/content-blocks/GridBlock/CarouselGrid/index.js
@@ -3,6 +3,7 @@ import PropTypes from "prop-types";
 import striptags from "striptags";
 import chunk from "lodash/chunk";
 import Tile from "@/atomic/Tile";
+import Loader from "@/atomic/Loader";
 import { Grid } from "@rubin-epo/epo-react-lib";
 import { normalizeItemData, useList } from "@/lib/utils";
 import * as Styled from "./styles";
@@ -19,7 +20,7 @@ function CarouselGrid({
   // get manually-curated data first
   const curatedItems = normalizeItemData(items);
 
-  const { data } = useList({
+  const { data, isLoading } = useList({
     limit,
     listTypeId,
     section,
@@ -64,6 +65,7 @@ function CarouselGrid({
     );
   }
 
+  if (isLoading) return <Loader speed="fast" isVisible />;
   if (!allItems?.length) return null;
 
   if (allItems.length <= 3)
--- a/components/dynamic/DataList/index.js
+++ b/components/dynamic/DataList/index.js
@@ -1,13 +1,16 @@
 import { useRouter } from "next/router";
 import PropTypes from "prop-types";
 import { useTranslation } from "react-i18next";
-import { Container } from "@rubin-epo/epo-react-lib";
+import { Container, Buttonish } from "@rubin-epo/epo-react-lib";
+import Loader from "@/atomic/Loader";
+import Pagination from "@/page/Pagination";
 import {
   usePathData,
   normalizePathData,
   useList,
   useGlobalData,
 } from "@/lib/utils";
+import * as Styled from "./styles";
 
 const DataList = ({
   children,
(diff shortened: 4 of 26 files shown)


components/atomic/Loader/index.js
```diff
@@ -0,0 +1,9 @@
+import * as Styled from "./styles";
+
+function Loader(props) {
+  return <Styled.Loader {...props} />;
+}
+
+Loader.displayName = "Atomic.Loader";
+
+export default Loader;
```
blnkt, line 9: Lil Styled version of the spinner from the UI lib.  Styles just center it horizontally and vertically in its parent.

components/atomic/Loader/styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+import { CircularLoader } from "@rubin-epo/epo-react-lib";
+
+export const Loader = styled(CircularLoader)`
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
blnkt, line 7: Styles just center it horizontally and vertically in its parent.

components/content-blocks/GridBlock/CarouselGrid/index.js
```diff
@@ -64,6 +65,7 @@ function CarouselGrid({
     );
   }
 
+  if (isLoading) return <Loader speed="fast" isVisible />;
```
blnkt, line 68: The useList data fetching hook was already returning a results object that included the loading state, so in a lot of cases adding the loading spinner places was just as simple as exposing that state and doing something with it.

components/dynamic/DataList/index.js
```diff
@@ -49,29 +55,54 @@ const DataList = ({
...
     );
+  }
 
-  return children(data);
+  if (isError) return null;
+
+  return (
+    <>
+      <Container width={width}>
+        {header && <Styled.Header>{header}</Styled.Header>}
```
blnkt, line 90: This header was common to every component that used `DataList` so I refactored into the `DataList` component itself

components/dynamic/DataList/index.js
```diff
@@ -49,29 +55,54 @@ const DataList = ({
...
+        {!isLoading && data && renderData()}
+        {footerButton && (
+          <Styled.Footer>
+            <Buttonish
+              isBlock={true}
+              text={footerButton.text}
+              url={`/${footerButton.uri}`}
+            />
+          </Styled.Footer>
+        )}
```
blnkt, line 101: This footer was common to every component that used DataList so I refactored into the DataList component itself

components/dynamic/DataList/index.js
```diff
@@ -49,29 +55,54 @@ const DataList = ({
...
+      <Pagination limit={limit} offset={offset} page={page} total={total} />
     );
+  }
 
-  return children(data);
+  if (isError) return null;
+
+  return (
+    <>
+      <Container width={width}>
```
blnkt, line 89: This container was common to every component that used DataList so I refactored into the DataList component itself

components/dynamic/DataList/index.js
```diff
@@ -49,29 +55,54 @@ const DataList = ({
...
+        {header && <Styled.Header>{header}</Styled.Header>}
+        {isLoading && <Loader speed="fast" isVisible />}
+        {!isLoading && data && renderData()}
```
blnkt, line 92: With all of the common elements refactored out of the components that use `DataList`, and into `DataList` itself, was able to also toggle children and Loader visibility here, instead of import a loader into every component where `DataList` is used.  I think there's actually even more opportunity for refactoring these dynamic List type components, as well as also the GridBlock components.  There seems to be quite a lot of commonality about the items in the List/Gird, such that some of the slight differences might be better controlled with props instead of discrete components for every use case.

components/dynamic/DataList/styles.js
```diff
@@ -0,0 +1,12 @@
+import styled from "styled-components";
+import { fluidScale } from "@/styles/globalStyles";
+
+export const Header = styled.h2`
```
blnkt, line 4: Moved inline styled component stuff into dedicated `styles.js` files wherever was fast and easy to do so in this refactor

lib/localeStrings/en.json
blnkt, line 1: localeStrings for loading description

pages/_document.js
```diff
@@ -2,8 +2,6 @@ import Document, { Html, Head, Main, NextScript } from "next/document";
 import { ServerStyleSheet } from "styled-components";
 import { getSiteString } from "@/lib/utils";
 
-const GOOGLE_APP_ID = process.env.NEXT_PUBLIC_GOOGLE_APP_ID;
```
blnkt, line 5: Not sure if this is suppose to be here or not....but sure doesn't seem to be in use in this file, so I removed

components/atomic/Loader/styles.js
alexgoff: do you think this change should go upstream to the library?

components/dynamic/DataList/index.js
alexgoff: good de-duplication

components/atomic/Loader/styles.js
blnkt: Probably?  Feels like a pretty sensible default styling for the spinner.  And wouldn't be much to override if ever wanted to use otherwise.

PR #332, opened 2023-06-13
[F] hourly precipitation widget

alexgoff:
Hourly precipitation widget, extracted a lot of the code from the windspeed one and formed it into an "hourly data" primitive with precipitation and windspeed as two patterns. Also moved a bunch of formatters out into a reusable helper module


blnkt (approved):
I like the "patterns" approach to this.  I think this is the kind of change I should next bring to the very similar datalists variants.  I also particularly appreciate refactoring the formatters into helper methods.

PR #333, opened 2023-06-14
[F] SummitData provider and basic flux query

alexgoff:


components/dynamic/SummitData/Weather/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+const Weather = () => {
+  const { tempUnit, windspeedUnit } = useWeatherUnit();
+  const { data, loading } = useSummitData();
+
+  if (loading || !data) return null;
+
+  const { temperature0: temperature, relativeHumidity, windSpeed } = data;
+
+  const temperatureData = {
```
alexgoff, line 16: data that will need to be converted is grouped together

lib/api/efd.js
```diff
@@ -0,0 +1,37 @@
...
+
+  const influxDB = new InfluxDB({ url, token });
+  const queryApi = influxDB.getQueryApi("");
+
+  return new Promise((resolve, reject) => {
+    let data = {};
+    queryApi.queryRows(
+      flux`from(bucket: "${bucket}")
+      |> range(start: -60s)
+      |> filter(fn: (r) => r["_measurement"] == "lsst.sal.ESS.temperature" or r["_measurement"] == "lsst.sal.ESS.relativeHumidity" or r["_measurement"] == "lsst.sal.ESS.airFlow")
```
alexgoff, line 16: trimmed down to just the current data that we need: temperature, humidity, wind, and need to get precipitation

contexts/SummitData.js
```diff
@@ -0,0 +1,92 @@
...
+        })
+        .catch((e) => {
+          reject(e);
+        });
+    });
+
+  useEffect(() => {
+    setLoading(true);
+
+    Promise.allSettled([
```
alexgoff, line 58: use Promise.allSettled so that one single failure won't disrupt the entire parallel fetch and we can keep it in `loading` status while they all catch up

components/dynamic/SummitData/Weather/index.js
```diff
@@ -0,0 +1,51 @@
...
+import TemperatureCurrent from "@/components/widgets/TemperatureCurrent";
+import { useSummitData } from "@/contexts/SummitData";
+import { defaultUnits, useWeatherUnit } from "@/contexts/WeatherUnit";
+import convert from "convert";
+
+const Weather = () => {
+  const { tempUnit, windspeedUnit } = useWeatherUnit();
+  const { currentData: data, loading = true } = useSummitData();
+
+  if (loading && !data) return <Loader isVisible={true} />;
```
alexgoff, line 12: since `loading` won't go to `false` until all complete, we'll need to check in each container if the provider is loading and the data hasn't been set yet. Or we could go to individualized loading statuses but I think this is ok as-is

PR #336, opened 2023-06-21
[F] SummitStatus modal + weather widgets setup

alexgoff:
This adds the modal that will contain all "detail" sections on the summit dashboard along with an implementation of the weather widgets.

There is one other major change included here, the weather unit provider has been refactored so that the unit localization selector does not need to have props and callbacks passed up and down through the individual sections and their modals. I have also trimmed down the data queries for the weather data to improve load times.

### DIFF
--- a/components/content-blocks/SummitStatus/index.js
+++ b/components/content-blocks/SummitStatus/index.js
@@ -1,155 +1,22 @@
 import PropTypes from "prop-types";
-import { useState, Suspense } from "react";
 import { Container } from "@rubin-epo/epo-react-lib";
-import { useTranslation } from "react-i18next";
-import WeatherUnitContext from "@/contexts/WeatherUnit";
+import { WeatherUnitProvider } from "@/contexts/WeatherUnit";
 import UnitLocalization from "@/components/layout/UnitLocalization";
 import WidgetGrid from "@/components/layout/WidgetGrid";
-import WidgetPreview from "@/components/layout/WidgetPreview";
-<<<<<<< HEAD
-=======
-import TemperatureCurrent from "@/components/widgets/CurrentData/patterns/Temperature";
-import TemperatureHistoric from "@/components/widgets/TemperatureHistoric";
-import WindspeedHourly from "@/components/widgets/HourlyData/patterns/Windspeed";
-import PrecipitationHourly from "@/components/widgets/HourlyData/patterns/Precipitation";
->>>>>>> ebb054c ([F] SummitData provider)
 import { SummitDataProvider } from "@/contexts/SummitData";
 import Weather from "@/components/dynamic/SummitData/Weather";
 
 const SummitStatus = ({ summitStatusLayout, widgetPreviews = [] }) => {
-  /** this logic should be changed to useRouter after i18n refactor */
-  const {
-    i18n: { language = "en" },
-  } = useTranslation();
-  const [windspeedUnit, setWindspeedUnit] = useState(
-    language === "en" ? "NM" : "m"
-  );
-  const [tempUnit, setTempUnit] = useState(
-    language === "en" ? "fahrenheit" : "celsius"
-  );
-
-  const mockWindspeedData = [
-    { windspeed: 5.2342342, direction: 120 },
-    { windspeed: 4.7, direction: 112 },
-    { windspeed: 6.1, direction: 98 },
-    { windspeed: 3.8, direction: 105 },
-    { windspeed: 5.5, direction: 117 },
-    { windspeed: 7.2, direction: 125 },
-    { windspeed: 4.3, direction: 103 },
-    { windspeed: 6.8, direction: 108 },
-    { windspeed: 5.14343, direction: 115 },
-    { windspeed: 4.9, direction: 98 },
-    { windspeed: 5.7, direction: 105 },
-    { windspeed: 6.5, direction: 114 },
-    { windspeed: 3.9, direction: 99 },
-    { windspeed: 4.1, direction: 102 },
-    { windspeed: 5.3, direction: 108 },
-    { windspeed: 6.3, direction: 116 },
-    {},
-    {},
-    {},
-    {},
-    {},
-    {},
-    {},
-    {},
-  ];
-
-  const mockPrecipitationData = [
-    { probability: 0.0 },
-    { probability: 0.0 },
-    { probability: 0.0 },
-    { probability: 0.0 },
-    { probability: 0.0 },
-    { probability: 0.0 },
-    { probability: 0.0 },
-    { probability: 0.0 },
-    { probability: 0.01 },
-    { probability: 0.02 },
-    { probability: 0.03 },
-    { probability: 0.04 },
-    { probability: 0.05 },
-    { probability: 0.03 },
-    { probability: 0.02 },
-    { probability: 0.1 },
-    { probability: 0.04 },
-    { probability: 0.03 },
-    { probability: 0.0 },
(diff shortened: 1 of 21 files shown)


components/content-blocks/SummitStatus/index.js
```diff
@@ -1,111 +1,22 @@
...
-
   return (
     <Container bgColor="neutral95" width="wide" paddingSize="small">
-      <WeatherUnitContext.Provider value={{ tempUnit, windspeedUnit }}>
-        <UnitLocalization
-          {...{ tempUnit, windspeedUnit }}
-          onTempChangeCallback={(value) => setTempUnit(value)}
-          onWindChangeCallback={(value) => setWindspeedUnit(value)}
-        />
+      <WeatherUnitProvider>
```
alexgoff, line 12: pulled out all of the state setting, etc. in this component. It is layout only now.

components/dynamic/SummitData/Weather/Current/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+  if (loading && !currentData)
+    return (
+      <WidgetSection {...sectionProps}>
+        <Loader isVisible={true} />
+      </WidgetSection>
+    );
+
+  const { temperature0, windspeed, relativeHumidity } = currentData;
+  const temperature = convertTemperature(temperature0, tempUnit);
```
alexgoff, line 31: each of the sections of the weather widgets has it's own container: current, hourly, and daily data. They are roughly the same, each pulls out it's data from the `useSummitData()` hook and then puts it into components.

This brings me to a point I want to discuss. Should the summit data provider handle the pre-processing for the data it retrieves from the EFD? Right now it just stores the raw values and the widget containers handle processing which is a combo of morphing into objects that the display components recognize and converting to selected temperature/windspeed units.

I think converting to end user temp/wind units is over-concerned and complicated, but what about stuff like setting up the objects themselves? Here's a few things I think could be pre-processed:

- mapping to recognizable properties (temperature0 -> temperature, speed_mean -> windspeed, direction_mean -> direction)
- slicing off extra data, for example if you query backwards 7 days you end up with 8 results because it also queries today from midnight -> present.
- putting timestamps into date objects

components/widgets/CurrentData/patterns/Windspeed/index.js
```diff
@@ -18,7 +19,7 @@ const WindspeedCurrent = ({ windspeed = 0, unit = "m" }) => {
       <Styled.Value $variant="large">
         {new Intl.NumberFormat(language, {
           style: "decimal",
-          maximumFractionDigits: 0,
+          maximumFractionDigits: unit === defaultWindspeedUnit ? 1 : 0,
```
alexgoff, line 22: this needs to match the hourly data display and show one decimal for m/s

contexts/WeatherUnit.js
```diff
@@ -1,14 +1,64 @@
...
+      return { ...state, windspeedUnit: "mi" };
+    case "m":
+      return { ...state, windspeedUnit: "m" };
+    default:
+      throw new Error(`Unhandled action type: ${type}`);
+  }
 };
 
-const WeatherUnitContext = createContext(defaultUnits);
+export const WeatherUnitProvider = ({ children }) => {
```
alexgoff, line 29: did a lot of work on the weather unit provider. Originally the provider was storing values set by the state in the summit status block, but with all of the modals having a unit localization control that means a lot of passing callbacks and props around. 

Instead I opted to move the state management to the provider itself and use a reducer to track the current units. Any instance of the unit localization component will receive the state and a dispatch function that can be used to set the unit and keep callbacks + re-renders to a minimum

contexts/WeatherUnit.js
```diff
@@ -1,14 +1,64 @@
...
 
 export function useWeatherUnit() {
-  return useContext(WeatherUnitContext);
+  const context = useContext(WeatherUnitContext);
+
+  if (context === undefined) {
+    throw new Error("useWeatherUnit must be used within a WeatherUnitProvider");
+  }
+
+  return context;
```
alexgoff, line 56: since UnitLocalization will be directly using this context instead of rendering from props, added an error in case the context is unavailable

lib/api/queries/weather.js
```diff
@@ -9,7 +9,7 @@ export const currentWeather = (bucket) => flux`from(bucket: "${bucket}")
   |> group()
   |> pivot(rowKey:["_pivoter"], columnKey: ["_field"], valueColumn: "_value")
   |> drop(columns: ["_pivoter"])
-  |> rename(columns: {direction: "windDirection", speed: "windSpeed"})
+  |> rename(columns: {direction: "windDirection", speed: "windspeed"})
```
alexgoff, line 12: standardized to use `windspeed` everywhere

lib/api/queries/weather.js
```diff
@@ -44,23 +44,19 @@ export const dailyWeather = (bucket) => {
   date.setHours(0, 0, 0, 0);
   return flux`data = from(bucket: "${bucket}")
       |> range(start: ${date}, stop: now())
-      |> filter(fn: (r) => r["_measurement"] == "lsst.sal.ESS.pressure" or r["_measurement"] == "lsst.sal.ESS.dewPoint" or r["_measurement"] == "lsst.sal.ESS.temperature" or r["_measurement"] == "lsst.sal.ESS.relativeHumidity" or r["_measurement"] == "lsst.sal.ESS.airFlow")
-      |> filter(fn: (r) => r["_field"] == "relativeHumidity" or r["_field"] == "dewPoint" or r["_field"] == "temperature0" or r["_field"] == "speed" or r["_field"] == "pressure0" or r["_field"] == "direction")
+      |> filter(fn: (r) => r["_measurement"] == "lsst.sal.ESS.temperature")
```
alexgoff, line 47: trimmed down some of the queries for speed, temperature is our only per-day widget so we don't need to fetch anything else and it only shows high/low so the mean isn't necessary

components/dynamic/SummitData/Weather/Current/index.js
```diff
@@ -0,0 +1,47 @@
...
+  const [isOpen, setOpen] = useState(true);
+  const [{ tempUnit, windspeedUnit }] = useWeatherUnit();
+  const { currentData, loading = true } = useSummitData();
+  const sectionProps = {
+    title: t("summit_dashboard.sections.weather.current"),
+    onToggleCallback: (value) => setOpen(value),
+    isOpen,
+  };
+
+  if (loading || !currentData)
```
alexgoff, line 23: super rudimentary undefined protection, it'll just stick in loading state forever. We should consult José on what kind of error status to put the UI in. Since we're going to be calling from a bunch of data sources I think having a robust error handling solution would be wise

contexts/SummitData.js
```diff
@@ -9,12 +16,31 @@ import queryEfd from "@/lib/api/efd";
...
+  dailyData: true,
+};
+
 export const SummitDataProvider = ({ children }) => {
   const [currentData, setCurrentData] = useState();
   const [hourlyData, setHourlyData] = useState();
   const [dailyData, setDailyData] = useState();
   const [error, setError] = useState(false);
-  const [loading, setLoading] = useState(true);
+  const [loading, dispatch] = useReducer(loadingReducer, defaultLoadState);
```
alexgoff, line 43: tracking the individual loading states of the datasets, trying to make them appear at the right times and not access data when it's undefined was too tricky with a single global loading state

blnkt (approved):

PR #338, opened 2023-06-26
[EPO-8239] EFD secrets are secret

blnkt:
A bit of an add on and resulting refactor.  As of writing this comment, the USDF is down for scheduled maintainence, so a little tricky to 100% know this is all working, but suffice to say the goals of this change were thus:
- move the use of the EFD credentials serverside so they're not exposed to the browser/clever users

And then, while I was in there, [based on some of the examples of this kind of thing I was seeing](https://www.smashingmagazine.com/2021/12/protect-api-key-production-nextjs-api-route/) it seemed like I could also incorporate SWR, and that it would do some/all of the statey promisey stuff for us?  But that's both the part that's hard for me to understand without actually being able to reach the USDF EFD, and also the part I am the least knowledgable in. I very well may have taken this part too far and dissolved some of the custom statey promisey stuff you had and that other components were very much depending on.  So take a look and let me know what you think.

### DIFF
--- a/contexts/SummitData.js
+++ b/contexts/SummitData.js
@@ -1,112 +1,53 @@
 import {
   createContext,
   useContext,
-  useEffect,
   useMemo,
-  useState,
-  useReducer,
 } from "react";
-import {
-  currentWeather,
-  dailyWeather,
-  hourlyWeather,
-} from "@/lib/api/queries/weather";
 import PropTypes from "prop-types";
-import queryEfd from "@/lib/api/efd";
+import useEfd from "@/lib/api/efd";
 
 export const SummitDataContext = createContext({});
 
-const loadingReducer = (state, action) => {
-  const { type, dataSet } = action;
-
-  switch (type) {
-    case "complete":
-      return { ...state, [dataSet]: false };
-    case "loading":
-      return { ...state, [dataSet]: true };
-    default:
-      throw new Error(`Unhandled action type: ${type}`);
-  }
-};
-
-const defaultLoadState = {
-  currentData: true,
-  hourlyData: true,
-  dailyData: true,
-};
-
 export const SummitDataProvider = ({ children }) => {
-  const [currentData, setCurrentData] = useState();
-  const [hourlyData, setHourlyData] = useState();
-  const [dailyData, setDailyData] = useState();
-  const [error, setError] = useState(false);
-  const [loading, dispatch] = useReducer(loadingReducer, defaultLoadState);
-
-  const fetchCurrentWeather = () =>
-    new Promise((resolve, reject) => {
-      queryEfd(currentWeather)
-        .then((value) => {
-          setCurrentData(...value);
-          resolve(...value);
-        })
-        .catch((e) => {
-          reject(e);
-        })
-        .finally(() => {
-          dispatch({ type: "complete", dataSet: "currentData" });
-        });
-    });
-
-  const fetchHourlyWeather = () =>
-    new Promise((resolve, reject) => {
-      queryEfd(hourlyWeather)
-        .then((value) => {
-          setHourlyData(value);
-          resolve(value);
-        })
-        .catch((e) => {
-          reject(e);
-        })
-        .finally(() => {
-          dispatch({ type: "complete", dataSet: "hourlyData" });
-        });
-    });
-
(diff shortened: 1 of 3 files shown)


pages/api/efd/[type].js
Comment (blnkt) [inline pages/api/efd/[type].js:1]: new internal API endpoint to hit for EFD data.  You pass it the type (current, hourly, daily) and it returns the influxdb response (data or error).  Next.js conventionally, because all of the files in the `pages/api` dir are only run serverside, we can therefore safely use our EFD env var credentials without exposing them to the browser.  This feels pertinent in this iteration of the Summit Status Dashboard because I think it someone scraped theour credentials out of the devtools network panel or whatever, they could have unrestricted read only access to the USDF EFD, which feels like something we don't wanna open the door to.

lib/api/efd.js
```diff
@@ -1,25 +1,13 @@
...
-      },
-      error(error) {
-        reject(error);
-      },
-      complete() {
-        resolve(data);
-      },
-    });
-  });
+  return {
```
blnkt, line 8: Becuase the env vars and the infuxing all lives in the `pages/api/efd` logic, this function is now only responsible for making the fetch to our internal API endpoint, and based on [some patterns I was reading about ](https://nextjs.org/docs/pages/building-your-application/data-fetching/client-side#client-side-data-fetching-with-swr), it seemed like I could use SWR here?

contexts/SummitData.js
```diff
@@ -1,112 +1,53 @@
...
+  const {
+    data: hourlyData,
+    isLoading: hourlyIsLoading,
+    isError: hourlyIsError,
+  } = useEfd("hourly");
+  const {
+    data: dailyData,
+    isLoading: dailyIsLoading,
+    isError: dailyIsError,
+  } = useEfd("daily");
```
blnkt, line 26: the `pages/api/efd` is now responsible for putting together the influxDB queries, based on the `type` that's specified via the internal API url that is fetched from `lib/api/efd`, so I think all of the promisey state handling fetching stuff can go away, and all happens elsewhere now, getting tucked behind this u`useEfd` hook.  Maybe?  It's honestly a bit hard to tell if I went about his last bit the correct way because the USDF is under scheduled maintenance so I can't yet fully test.

contexts/SummitData.js
```diff
@@ -1,112 +1,53 @@
...
 
   const value = useMemo(
     () => ({
       currentData,
       hourlyData,
       dailyData,
-      loading,
-      error,
+      error: currentIsError || hourlyIsError || dailyIsError,
```
blnkt, line 33: error is true if any of the fetches returned an error

contexts/SummitData.js
```diff
@@ -1,112 +1,53 @@
...
+      loading: {
+        currentData: currentIsLoading,
+        hourlyData: hourlyIsLoading,
+        dailyData: dailyIsLoading,
+      },
```
blnkt, line 38: retaining the loading status per fetch approach

pages/api/efd/[type].js
```diff
@@ -29,12 +29,10 @@ export default function handler(req, res) {
...
     },
     error(error) {
-      // reject(error);
-      res.status(200).json({ error })
+      return res.status(405).json(error);
     },
     complete() {
-      // resolve(data);
-      res.status(200).json({ data })
+      return res.status(200).json(type === "current" ? data[0] : data);
```
Comment (blnkt) [inline pages/api/efd/[type].js]: I was doing 2 dumb things:

1. I was returning objects that looked like `{ data: { data: [stuff, and, things] } }` when it shoulda been `{ data: [stuff, and, things] }`
2. everywhere `currentData` was being used expected a single result instead of an array of results, so I adjusted here instead of downstream.  That is, I suppose a different way to approach could be to have all `summitData` consumers all always expect an array of data.  Up to @alexgoff which approach is best fit.

lib/api/efd.js
```diff
@@ -2,7 +2,7 @@ import useSWR from "swr";
 
 const fetcher = (url) => fetch(url).then((res) => res.json());
 
-export default async function useEfd(type) {
```
blnkt: useSWR is doing the promisey stuff for us now so `useEfd` should no longer be `async`

alexgoff (approved):

PR #347, opened 2023-07-17
[F] Moonrise/moonset forecast

alexgoff:
Moonrise and moonset forecast widget, the biggest change here is that I have pulled in the `suncalc` library to address timezone issues.

**Testing**

Moonrise and moonset times can be verified against this site: https://www.timeanddate.com/moon/chile/la-serena, we expect a difference of a few minutes since the observatory 

image


helpers/formatters.js
```diff
@@ -40,3 +40,21 @@ export const formatTime = (value, locale = "en", options = {}) => {
...
+  const today = new Date();
+  const result = new Date(today);
+  const diff = day - today.getDay();
+
+  result.setDate(today.getDate() + diff);
+
+  return result.toLocaleDateString(locale, { weekday: "long" });
+};
+
+export const formatAngle = (angle = 0, locale = "en") =>
```
alexgoff, line 54: extract `formatDayName` and `formatAngle` into formatters

lib/suncalc.js
alexgoff, line 1: this is the suncalc lib + `tz-lookup` to assist with getting the correct timezone for a lat/long

lib/suncalc.js
```diff
@@ -0,0 +1,358 @@
...
+
+function hoursLater(date, h) {
+  return new Date(date.valueOf() + (h * dayMs) / 24);
+}
+
+// calculations for moon rise/set times are based on http://www.stargazing.net/kepler/moonrise.html article
+
+export function getMoonTimes(date, lat, lng, inUTC = false) {
+  const t = new Date(date);
+  const tz = tzlookup(lat, lng);
```
alexgoff, line 294: here is the largest change I made, originally it was 

```javascript
if (inUTC) t.setUTCHours(0,0,0,0);
else t.setHours(0,0,0,0)
```
which would always set the time to the users system midnight and then try to find a moonrise and moonset in the next 24 hours. This would produce incorrect rises/sets for each day since the local timezone is different than user timezone. Now it checks for if the time is in UTC and if not calculates the local timezone offset from UTC and uses that to set midnight.

PR #364, opened 2023-08-09
[C] SVG cross browser compat

alexgoff:
Safari does not support several modern SVG standards so some cross browser work is required to get the same presentation across all browsers. The biggest change is getting rid of the `textPath` approach on the radial meter and changing `transform` to CSS transforms


components/atomic/RadialMeter/index.js
```diff
@@ -47,31 +44,38 @@ const RadialMeter = ({ min, max, value, text, step = 1, labelledById }) => {
...
-            y2={centerY}
+          <g
             transform-origin={origin}
-            transform={`rotate(${i * degreeStep + baseRotation})`}
-          />
+            style={{
+              transform: `rotate(${i * degreeStep + baseRotation}deg)`,
+            }}
+            key={v}
+          >
```
alexgoff, line 64: instead of moving the text along a path, I am attaching the text to it's corresponding tick and offsetting then rotating it from the center

components/charts/XAxis/index.js
```diff
@@ -33,7 +33,7 @@ const XAxis = ({
       {tickMap.map(({ label, position }, i) => (
         <g role="listitem" key={`${label}-${i}`}>
           <Styled.Tick x1={position} x2={position} y1={y} y2={y + tickHeight} />
-          <Styled.TickLabel x={position} y={y + tickHeight * 2}>
+          <Styled.TickLabel x={position} y={y + tickHeight * 3}>
```
alexgoff, line 36: increased the offset of the X axis labels instead of using dominantBaseline, since Safari does not support it

PR #384, opened 2023-12-22
[C] use original URL for GIF

alexgoff:
Canto's preview URL's are always in JPG format, this breaks GIF formats on smaller screen sizes. Short circuit this behavior by adding filetype to the asset query and using the original URL instead.


blnkt (approved):
Sorry I totally spaced this one.  But finally tested and indeed GIFs work as expected now.  also tested across components using the same field for not GIFs and works as before/expected.

PR #425, opened 2024-04-09
3 level desktop nav

blnkt:
Must test with https://github.com/lsst-epo/rubin-obs-api/pull/224

Introduces a sense of a 3rd level to the Header Nav.  Design is not quite consistent with Jose's original vision as implementing a non-accordion 3rd level in the desktop nav proved to be much more involved than anticipated.  Could circle back on that in the future, but was eager to release this work as-is so that Kristen and Steph may begin reorganizing the pagetree ASAP.  In addition to new styles and functionality to support a 3rd level, there are also many small stylistic updates to the header and the nav in general (all as seen in Jose's latest design).

### What to Test
On Desktop:
- If a nav drop down is open it should remain open, and the header should remain visible on scroll

On mobile:
- If a mobile nav is open it should remain open, and the header should remain visible on scroll

On Both
- If 2nd or 3rd level nav items are visible, they should be hidden after you switch to another set of 2nd or 3rd level nav items
- If you switch language nav should close
- Nav should visually match (approximately) design, with the exception that we use the mobile accordion style for 3rd level nav items on mobile and desktop

### DIFF
--- a/components/global/Header/NavItemWithChildren.js
+++ b/components/global/Header/NavItemWithChildren.js
@@ -9,12 +9,15 @@ import internalLinkShape, { internalLinkInternalShape } from "@/shapes/link";
 export default function NavItemWithChildren({
   id,
   active,
+  parentActive,
   title,
   uri,
   childItems,
   onToggleClick,
   onEsc,
   theme,
+  baseClassName = "c-nav-list",
+  level = 2,
 }) {
   const ref = useRef(null);
 
@@ -30,24 +33,30 @@ export default function NavItemWithChildren({
   const childrenWithParent = [parent].concat(childItems);
 
   return (
-    <div ref={ref}>
+    <div ref={ref} className={`${baseClassName}__item-inner`}>
       <button
         onClick={() => onToggleClick(id)}
         aria-expanded={active}
         aria-haspopup
         className={classNames({
-          "c-nav-list__link": true,
-          "c-nav-list__link--is-active": active,
-          [`c-nav-list__link--${theme}`]: !!theme,
+          [`${baseClassName}__link`]: true,
+          [`${baseClassName}__link--is-active`]: active,
+          [`${baseClassName}__link--${theme}`]: !!theme,
         })}
       >
-        <IconComposer icon="ChevronThin" className="c-nav-list__link-icon" />
-        <span className="c-nav-list__link-text">{title}</span>
+        <IconComposer
+          icon="ChevronThin"
+          className={`${baseClassName}__link-icon`}
+        />
+        <span className={`${baseClassName}__link-text`}>{title}</span>
       </button>
       <Subnavigation
         items={childrenWithParent}
-        active={active}
+        active={level === 3 ? parentActive && active : active}
         onClick={onEsc}
+        theme={theme}
+        level={3}
+        baseClassName={level === 3 ? baseClassName : "c-subnav-list"}
       />
     </div>
   );
@@ -59,8 +68,11 @@ NavItemWithChildren.displayName =
 NavItemWithChildren.propTypes = {
   ...internalLinkInternalShape,
   active: PropTypes.bool,
+  parentActive: PropTypes.bool,
   childItems: PropTypes.arrayOf(internalLinkShape),
   onToggleClick: PropTypes.func.isRequired,
   onEsc: PropTypes.func.isRequired,
   theme: PropTypes.oneOf(["desktop", "mobile"]),
+  baseClassName: PropTypes.string,
+  level: PropTypes.number,
 };
--- a/components/global/Header/Navigation.js
+++ b/components/global/Header/Navigation.js
@@ -1,4 +1,4 @@
-import { useState, useRef } from "react";
+import { useEffect, useState, useRef } from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
 import { useTranslation } from "react-i18next";
@@ -13,6 +13,7 @@ export default function Navigation({
   items,
   userProfilePage,
   theme,
+  isOpenSetter,
   mobileActive,
   mobileSetter,
 }) {
@@ -27,12 +28,17 @@ export default function Navigation({
 
   useClickEvent(handleClick);
 
+  useEffect(() => {
+    if (isOpenSetter) isOpenSetter(!!active);
+  }, [mobileSetter, isOpenSetter, active]);
+
   function handleClick(e) {
     const isLink =
       e.target.nodeName === "A" || e.target.parentElement?.nodeName === "A";
     const isMobileLangSelect = e.target.id === mobileLangSelectId;
     if (mobileSetter && mobileActive && (isLink || isMobileLangSelect)) {
       mobileSetter(false);
+      setActive(null);
     }
     if (navList?.current?.contains(e.target)) return;
     setActive(null);
@@ -159,6 +165,7 @@ Navigation.propTypes = {
   items: PropTypes.arrayOf(internalLinkWithChildrenShape),
   userProfilePage: PropTypes.object,
   theme: PropTypes.oneOf(["desktop", "mobile"]),
+  isOpenSetter: PropTypes.func,
   mobileActive: PropTypes.bool,
   mobileSetter: PropTypes.func,
 };
--- a/components/global/Header/Navigation.js
+++ b/components/global/Header/Navigation.js
@@ -1,4 +1,4 @@
-import { useEffect, useState, useRef } from "react";
+import { useState, useRef } from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
 import { useTranslation } from "react-i18next";
@@ -13,7 +13,7 @@ export default function Navigation({
   items,
   userProfilePage,
   theme,
-  isOpenSetter,
+  desktopSetter,
   mobileActive,
   mobileSetter,
 }) {
@@ -28,10 +28,6 @@ export default function Navigation({
 
   useClickEvent(handleClick);
 
-  useEffect(() => {
-    if (isOpenSetter) isOpenSetter(!!active);
-  }, [mobileSetter, isOpenSetter, active]);
-
   function handleClick(e) {
     const isLink =
       e.target.nodeName === "A" || e.target.parentElement?.nodeName === "A";
@@ -42,15 +38,17 @@ export default function Navigation({
     }
     if (navList?.current?.contains(e.target)) return;
     setActive(null);
+    if (desktopSetter) desktopSetter(false);
   }
 
   function handleToggleClick(id) {
     setActive((prevActive) => (prevActive === id ? null : id));
+    if (desktopSetter) desktopSetter(true);
   }
 
   function handleSignOut() {
     setActive(null);
-    mobileSetter(false);
+    if (mobileSetter) mobileSetter(false);
     signOut();
   }
 
@@ -165,7 +163,7 @@ Navigation.propTypes = {
   items: PropTypes.arrayOf(internalLinkWithChildrenShape),
   userProfilePage: PropTypes.object,
   theme: PropTypes.oneOf(["desktop", "mobile"]),
-  isOpenSetter: PropTypes.func,
+  desktopSetter: PropTypes.func,
   mobileActive: PropTypes.bool,
   mobileSetter: PropTypes.func,
 };
--- a/components/global/Header/Subnavigation.js
+++ b/components/global/Header/Subnavigation.js
@@ -1,34 +1,75 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */
+/*
+ * Ignore false positives (key event exists on Navigation component; role attribute not needed
+ * since Link component validly handles href and window history, click just closes subnav)
+ */
+import { useState, useEffect } from "react";
 import PropTypes from "prop-types";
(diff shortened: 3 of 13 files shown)


components/global/Header/NavItemWithChildren.js
```diff
@@ -9,12 +9,15 @@ import internalLinkShape, { internalLinkInternalShape } from "@/shapes/link";
 export default function NavItemWithChildren({
   id,
   active,
+  parentActive,
```
blnkt, line 12: Introduced a sense of "is my parent nav item active?" in order to collapse sibling 3rd level accordions

components/global/Header/NavItemWithChildren.js
```diff
@@ -9,12 +9,15 @@ import internalLinkShape, { internalLinkInternalShape } from "@/shapes/link";
...
   title,
   uri,
   childItems,
   onToggleClick,
   onEsc,
   theme,
+  baseClassName = "c-nav-list",
+  level = 2,
```
blnkt, line 20: Introduced sense of "what level am I in the nav(s)" in order to recursively reuse the NavItemWithChildren and Subnavigation components

components/global/Header/NavItemWithChildren.js
```diff
@@ -30,24 +33,30 @@ export default function NavItemWithChildren({
   const childrenWithParent = [parent].concat(childItems);
 
   return (
-    <div ref={ref}>
+    <div ref={ref} className={`${baseClassName}__item-inner`}>
```
blnkt, line 36: `baseClassName` passed in to `NavItemWithChildren` and `Subnavigation`

components/global/Header/Navigation.js
```diff
@@ -27,12 +28,17 @@ export default function Navigation({
 
   useClickEvent(handleClick);
 
+  useEffect(() => {
+    if (isOpenSetter) isOpenSetter(!!active);
```
blnkt, line 32: `isOpen` is responsible for telling the `Header` not to hide on scroll if the nav is open.

components/global/Header/Subnavigation.js
```diff
@@ -1,34 +1,75 @@
...
+                tabIndex={active ? 0 : -1}
+                onClick={onClick}
+              >
+                {title}
+              </Link>
+            )}
+          </li>
+        );
+      })}
     </ul>
```
blnkt, line 73: recursiveness

components/global/Header/_navigation.scss
```diff
@@ -5,18 +5,11 @@ $_breakpoint: functions.header(layout-breakpoint);
 $_link-gap: 14px;
 
 .c-nav-list {
-  @include base.fluid-scale(
-    nav-width,
-    functions.header(nav-width-desktop),
-    functions.header(nav-width-mobile),
-    functions.break(tablet),
-    functions.break(mobile)
-  );
```
blnkt, line 14: width got more finicky with more levels so having this as a css var became more of a hindrance than a help

components/global/Header/_subsubnavigation.scss
```diff
@@ -0,0 +1,108 @@
+@use "abstracts/functions";
+
```
blnkt, line 2: new styles for 3rd level of the nav.  heavily based on the 2nd level styles with some important differences, principally padding around nav items and the accordion stuff

components/global/Header/_subsubnavigation.scss
```diff
@@ -0,0 +1,108 @@
...
+      transform: rotate(-90deg);
+    }
+
+    &-text {
+      max-width: calc(100% - 34px);
+      text-align: left;
+    }
+
+    &--is-active &-icon {
+      transform: rotate(90deg);
```
blnkt, line 105: I don't love rotating an icon just to get the starting orientation I need, but there was not caret with the correct orientation, and I was always gonna rotate it onclick anyhow so 🤷

components/global/Header/index.js
```diff
@@ -56,7 +57,7 @@ export default function Header({ navItems, userProfilePage }) {
     <header
       ref={ref}
       className={`c-global-header ${
-        visible || mobileNavActive ? "" : "invisible"
+        visible || mobileNavActive || isOpen ? "" : "invisible"
```
blnkt, line 60: keep the header visible if the nav is open

lib/api/fragments/global.js
```diff
@@ -5,10 +5,15 @@ fragment pageTreeFragment on EntryInterface {
   id
   title
   uri
-  children(type: ["not", "educatorPages"]) {
+  children(isVisible: true) {
```
blnkt, line 8: filtering on `isVisible` to only get the nav items that ought to be there.  Also deleted the filter on `educatorPages` because I couldn't think of any reason to not allow them in the nav to begin with.  Not sure what the previous rationale might have been for that exclusion but.....I think it's fine.  Access to them is restricted but I think there's no harm in allowing someone to reach those pages.

theme/styles/abstracts/_variables.scss
```diff
@@ -66,9 +66,9 @@ $header: (
   gap: 0,
   layout-breakpoint: 1625px,
   breakpoint-for-search-bar: 850px,
-  nav-link-padding-lateral: 0.8em,
-  nav-width-desktop: 300px,
-  nav-width-mobile: 225px,
+  nav-link-padding-lateral: 1em,
+  nav-width-desktop: 600px,
+  nav-width-mobile: 300px,
```
blnkt, line 71: updating sass vars to match updated design.  One design departure is Jose showed the 1st level mobile nav significantly narrower than the 2nd level.  It looked more balanced to me to keep the 1st and 2nd level the same width on most screen sizes, so that's what I did.

theme/styles/globalStyles.js
```diff
@@ -42,7 +42,6 @@ export const tokens = {
   red20: "#f2c3c0",
   red40: "#FF8489",
   BREAK_HEADER_LAYOUT: "1625px",
-  BREAK_HEADER_LAYOUT: "1500px",
```
blnkt, line 45: duplicate token that I think is totally erroneous so I deleted

components/global/Header/Subnavigation.js
```diff
@@ -1,34 +1,75 @@
...
+  items,
+  active,
+  onClick,
+  theme,
+  baseClassName = "c-subnav-list",
+  level = 2,
+}) {
+  const [activeSub, setActiveSub] = useState(null);
+
+  useEffect(() => {
```
alexgoff, line 23: does this need an effect? I see that it's there to prevent endless renders setting `activeSub` after `active` changes, but it looks like it could be removed, the only replacement necessary is in the render doing:

```
        const isActiveSub = active && id === activeSub;
```

otherwise we're splitting the state into two separate updates, which can feel sluggish

components/global/Header/_subnavigation.scss
```diff
@@ -6,18 +6,17 @@ $_bg-color: functions.sass-palette(neutral02);
 .c-subnav-list {
   position: absolute;
   z-index: -5;
-  width: var(--nav-width);
   max-width: 47vw;
   max-height: calc(100vh - var(--header-height));
   overflow: auto;
```
alexgoff, line 11: I'm not sure what the original purpose was, but with "auto" set, there is a flash of scrollbar when opening/closing the subnav. I tested unsetting it and it was ok, but there might be a use case I'm unaware of

components/global/Header/Navigation.js
```diff
@@ -27,12 +28,17 @@ export default function Navigation({
 
   useClickEvent(handleClick);
 
+  useEffect(() => {
```
alexgoff, line 31: can this effect be done at the same time as changing the `active` state? 

the current React recommendation is to do the state change + callback together, outside an effect: https://react.dev/learn/you-might-not-need-an-effect#notifying-parent-components-about-state-changes

alexgoff (changes requested):
One note about the styles that I think needs attention to avoid an unintentional flash of scrolling, then two notes about effects that are nice-to-check but functionality-wise everything is as described.

components/global/Header/Subnavigation.js
blnkt: good catch.  The management of like open state was something I circled back on after the fact, and you're right, in this case it is more practical to just use the value of the `active` prop rather than use it to trigger internal state change.

components/global/Header/_subnavigation.scss
blnkt: good catch.  I think you mean to have this comment on `_subsubnavigation.scss` (the new level), and you're totally right, because I'm transitioning `max-height` we get a flash of scrollbar during transition.  Luckily, for every stylistic idiosyncrasy there is [a css tricks](https://css-tricks.com/hide-scrollbars-during-an-animation/).  Never encountered this one before, but seems to be working as advertised where I've tested.

components/global/Header/Subnavigation.js
blnkt: Ooo.  Actually, I take it back.  I think this is the best (only?) way I know how to ensure 3rd level accordions are collapsed if their parent dropdown is closed.  Which I'm doing so that navs start out all neat and collapsed every time their parent dropdown is opened.

components/global/Header/Navigation.js
blnkt: removed `useEffect` and instead (as sparingly as appears reasonable) call my parent state setter function where/when the active nav id is set.

components/global/Header/Navigation.js
```diff
@@ -28,10 +28,6 @@ export default function Navigation({
 
   useClickEvent(handleClick);
 
-  useEffect(() => {
-    if (isOpenSetter) isOpenSetter(!!active);
-  }, [mobileSetter, isOpenSetter, active]);
-
```
blnkt: removed the use of `useEffect`

components/global/Header/Navigation.js
```diff
@@ -42,15 +38,17 @@ export default function Navigation({
     }
     if (navList?.current?.contains(e.target)) return;
     setActive(null);
+    if (desktopSetter) desktopSetter(false);
```
blnkt, line 41: Instead of `useEffect` in response to `active` changing, now just calling parent state setter (`desktopSetter`) where `setActive` is called.  Note, it is not called everywhere `setActive` is called, only when the value of `active` will impact the on-screen visibility of the desktop nav.

components/global/Header/Navigation.js
```diff
@@ -165,7 +163,7 @@ Navigation.propTypes = {
   items: PropTypes.arrayOf(internalLinkWithChildrenShape),
   userProfilePage: PropTypes.object,
   theme: PropTypes.oneOf(["desktop", "mobile"]),
-  isOpenSetter: PropTypes.func,
+  desktopSetter: PropTypes.func,
```
blnkt, line 166: renamed to `desktopSetter` to clarify that it is only impactful to the desktop nav.

alexgoff (approved):

components/global/Header/Navigation.js
```diff
@@ -42,15 +38,17 @@ export default function Navigation({
...
 
   function handleToggleClick(id) {
     setActive((prevActive) => (prevActive === id ? null : id));
+    if (desktopSetter) desktopSetter(true);
   }
 
   function handleSignOut() {
     setActive(null);
-    mobileSetter(false);
+    if (mobileSetter) mobileSetter(false);
```
blnkt, line 51: `mobileSetter` only exists in the mobile instance of the nav, and while it may be the case that here it always exists because of when `handleSignOut` is called, elsewhere I saw `mobileSetter` called conditional on its own existence, so decided to do the same here. elsewhere I saw

components/global/Header/_subsubnavigation.scss
```diff
@@ -8,15 +8,23 @@ $_button-bg-color: functions.sass-palette(neutral02);
...
   color: $_color;
   visibility: hidden;
   transition: max-height 0.25s ease-out, visibility 0s 0.25s;
 
+  @keyframes hide-scroll {
+    from, to {
+      overflow: hidden;
+    }
+  }
+
```
blnkt, line 21: super interesting `overflow:hidden` flashy scroll bar hack.  Long and short of it is, `keyframes` also take the most precedence in stylesheet, and so can be effectively used as an override by making the animation duration the same as the `max-height` transition duration.

components/global/Header/index.js
```diff
@@ -57,7 +57,7 @@ export default function Header({ navItems, userProfilePage }) {
     <header
       ref={ref}
       className={`c-global-header ${
-        visible || mobileNavActive || isOpen ? "" : "invisible"
+        visible || mobileNavActive || desktopNavActive ? "" : "invisible"
```
blnkt, line 60: renamed to clarify it only impacts the desktop instance of the nav

theme/styles/abstracts/_variables.scss
```diff
@@ -67,7 +67,7 @@ $header: (
   layout-breakpoint: 1625px,
   breakpoint-for-search-bar: 850px,
   nav-link-padding-lateral: 1em,
-  nav-width-desktop: 600px,
+  nav-width-desktop: 450px,
```
blnkt, line 70: 600px width was a change I made to reflect design of desktop nav's dropdowns, but on closer inspection I realized that if the rightmost top level nav item had a dropdown, that on most desktop screen sizes it would overflow out the right side of the body container.

PR #432, opened 2024-04-24
People Content Block

blnkt:
**Related to**
https://github.com/lsst-epo/rubin-obs-api/pull/231

**What it Does**
Adds a People Entry type and a People Content Block available for page entries

**How to Test**
1. Create a handful of People entries, some with optional fields filled in, some without.
2. Add a People Content Block to a page
3. Save page entry and view on the frontend.  Expect no content block to render if no People have been added to the People Content Block
4. Add some Publications to the People Content Block
5. Save page entry and view on the frontend.  Expect content block to render now that it has People added


alexgoff:
With very long names, the grid items will grow to uneven sizes. The grid needs a column definition of
```css
grid-template-columns: repeat(3, minmax(0,1fr));
```

and the names need to be broken with

```css
word-break: break-word
```

image

alexgoff (changes requested):
With very long names, the grid items will grow to uneven sizes. The grid needs a column definition of
```css
grid-template-columns: repeat(3, minmax(0,1fr));
```

and the names need to be broken with

```css
word-break: break-word
```

image

PR #492, opened 2024-07-05
[C] Refactor i18n setup

alexgoff:
Refactoring the main site i18next setup to fall in line with the Investigations, and prepare for app router migration.

The main changes are:

- creating a new i18next instance with resources pulled in through an i18next backend, language detection, shared configs, and an unused server translation instance
- adding the next-intl/middleware to our current middleware to redirect users based on their browser locale and set a cookie that tracks locale.


ericdrosas87:
Screenshot 2024-07-22 at 12 01 08 PM

I wasn't able to get the _homepage_ or _about_ page to load in Spanish automatically even when I set Spanish to my preferred language in Chrome and set the Sensor tab to override my GPS to appear as if I'm in Mexico, but the Spanish translation toggle works fine with no errors.

ericdrosas87 (approved):

ericdrosas87 (approved):

PR #525, opened 2024-09-18
refactor: separate homepage from urisegments

### LINKED ISSUE #495: refactor: separate homepage from uriSegments
The `[[...uriSegments]].js` route houses all templates for all URI's. While other pages may not be immediate candidates for separation because their routing may change, the homepage is always expected to exist at the root directory of `/` or `[locale/` for non-English sites. 

Since the homepage also has unique conditions applied that make it different from other routes that use `...uriSegments`, there is a strong case for separating the homepage and it's separate logic into a new route with the following structure:

- [layout]
    - index.js <-- homepage
    - [[...uriSegments]].js <-- all other routes

This setup will enable better incremental migration into the app router as it simplifies the logic for each page.

alexgoff:
Separate the homepage from the `[...uriSegments]` catch-all since it has unique queries and URI structures. 

Side effect, refactor `getGlobalData` to query by current site rather than querying all sites and then filtering.

## Testing

Optimization benefits of separation:

### Original

- 100kb rendered, 381kb first load JS (per-page)
- 1.4mb total transferred
- 480kb JS

### Separated

- 2.75kb rendered, 352kb first load JS (for homepage)
- 1.3mb total transferred
- 397kb JS

Resolves #495

### DIFF
--- a/lib/api/global.js
+++ b/lib/api/global.js
@@ -10,7 +10,8 @@ import {
 } from "@/lib/api/fragments/global";
 import { categoriesFragment } from "@/lib/api/fragments/categories";
 import { userProfileFragment } from "@/lib/api/fragments/page";
-export async function getGlobalData() {
+
+export async function getGlobalData(site = "default") {
   const query = gql`
     ${linkFragment}
     ${siteInfoFragment}
@@ -19,10 +20,10 @@ export async function getGlobalData() {
     ${contactFormFragment}
     ${categoriesFragment}
     ${userProfileFragment}
-    {
+    query getGlobalData($site: [String]) {
       pageTree: entries(
         section: "pages"
-        site: "default"
+        site: $site
         level: 1
         isVisible: true
       ) {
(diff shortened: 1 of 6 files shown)


lib/api/global.js
```diff
@@ -19,42 +20,10 @@ export async function getGlobalData() {
     ${contactFormFragment}
     ${categoriesFragment}
     ${userProfileFragment}
-    {
+    query getGlobalData($site: [String]) {
```
alexgoff, line 23: previously all global data was being fetched twice and stored like:

```
data {
    global: {...},
    global_es: {...}`
}
```

This change adds the site as a param in the function and only retrieves globals for the current site, effectively halving the query size.

pages/[locale]/[...uriSegments]/index.js
Comment (alexgoff) [inline pages/[locale]/[...uriSegments]/index.js:1]: remove homepage references from uriSegments and change from an optional catch-all to a simple catch-all

pages/api/preview.js
```diff
@@ -47,14 +48,31 @@ const preview = async (request, response) => {
...
-  const redirectUri = res.entry.uri === CRAFT_HOMEPAGE_URI ? "" : res.entry.uri;
-  const redirect = `/${locale}/${redirectUri}`;
-  const cookiePath = `${site === "default" ? "" : `/${locale}`}/${redirectUri}`;
-
   // Enable Draft Mode by setting the cookie
   response.setDraftMode({ enable: true });
-  response.setPreviewData({ previewToken }, { path: cookiePath, maxAge: 120 });
-  response.redirect(redirect);
+
+  if (res.entry.uri === CRAFT_HOMEPAGE_URI) {
```
alexgoff, line 54: refine the preview URI structure to reflect that the homepage is now in it's own directory

ericdrosas87 (approved):

PR #528, opened 2024-09-20
refactor: add revalidatePath support to revalidate endpoint

### LINKED ISSUE #514: refactor: add revalidatePath support to `revalidate` endpoint
Add support for app directory’s [revalidatePath](https://nextjs.org/docs/app/building-your-application/caching#revalidatepath) API to our existing revalidation endpoint

alexgoff:
Adds an app directory Route Handler that performs revalidation to support to our existing `revalidate` endpoint. The existing `api/revalidate` endpoint will call this new `api/app-revalidate` endpoint with the same arguments.

Testing this is tricky, since we do not have any app directory routes, but can be done with the following:

- add an `app` folder with `[locale]` as a child folder, and `my-page` as a folder within `[locale]`. 
```
app
|- [locale]
  |- my-page
    |- page.jsx
```

- add a `page.jsx` to `my-page` that default exports a React component. Inside the component, fetch a random number from an API and display it.
- 
```jsx
const Page = async () => {
    const response = await fetch('http://www.randomnumberapi.com/api/v1.0/random?min=100&max=1000&count=1');
    const number = await response.json();

    return <h1>{number}</h1>
}
```

- In `app/api/app-revalidate/route.ts` add `revalidatePath('/[locale]/my-page', 'page')` after the existing `revalidatePath`
- Start a production build with `yarn bs`
- Visit `/es/my-page` and refresh a few times, the number should not change.
- Save an entry in CraftCMS, look for any console errors
- Visit `/es/my-page` again and the number should change

Resolves #514

### DIFF
--- a/app/api/app-revalidate/route.ts
+++ b/app/api/app-revalidate/route.ts
@@ -1,0 +1,37 @@
+import { revalidatePath } from "next/cache";
+import { NextRequest, NextResponse } from "next/server";
+
+const REVALIDATE_SECRET_TOKEN = process.env.CRAFT_REVALIDATE_SECRET_TOKEN;
+
+export async function GET(request: NextRequest): Promise<NextResponse> {
+  const uri = request.nextUrl.searchParams.get("uri");
+  const secret = request.nextUrl.searchParams.get("secret");
+
+  if (!uri) {
+    return NextResponse.json({
+      revalidated: false,
+      now: Date.now(),
+      message: "Missing path to revalidate",
+    });
+  }
+
+  if (secret !== REVALIDATE_SECRET_TOKEN) {
+    return NextResponse.json({
+      revalidated: false,
+      now: Date.now(),
+      message: "Invalid token",
+    });
+  }
+
+  if (uri) {
+    revalidatePath(`/[locale]/${uri}`, "page");
+
+    return NextResponse.json({ revalidated: true, now: Date.now() });
+  }
+
+  return NextResponse.json({
+    revalidated: false,
+    now: Date.now(),
+    message: "Error revalidating",
+  });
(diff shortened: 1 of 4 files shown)


pages/api/revalidate.js
```diff
@@ -22,6 +23,12 @@ async function handler(req, res) {
     return res.status(500).json({ error: `The parameter "uri" is required.` });
   }
 
+  try {
```
alexgoff, line 26: since we don't have an app directory setup, squash the error message that will throw when trying to revalidate path

pages/api/revalidate.js
```diff
@@ -23,6 +31,9 @@ async function handler(req, res) {
   }
 
   try {
+    const searchParams = new URLSearchParams(query);
+
+    await fetch(`${BASE_URL}/api/app-revalidate?${searchParams.toString()}`);
```
alexgoff, line 36: forward the request to `/app-revalidate`. Eventually this API will be deprecated and `/app-revalidate` will be renamed to `/revalidate` and assume the role as primary revalidator

app/api/app-revalidate/route.ts
```diff
@@ -0,0 +1,37 @@
+import { revalidatePath } from "next/cache";
+import { NextRequest, NextResponse } from "next/server";
+
+const REVALIDATE_SECRET_TOKEN = process.env.CRAFT_REVALIDATE_SECRET_TOKEN;
+
+export async function GET(request: NextRequest): Promise<NextResponse> {
```
alexgoff, line 6: direct clone of investigations revalidation setup

ericdrosas87 (approved):

PR #541, opened 2024-09-27
feat: app directory

alexgoff:
Port of the `[locale]` route segment from the Pages to App directory. 

### Testing

On homepage:

- Visual
    - Check side-by-side with production instance. Look for layout shifts, font changes, color changes, etc.
    - Check for hydration errors (will pop up)
    - Check for any obviously wrong behavior
- Metadata
    - Should have
        - Title (the title format will vary slightly from the existing homepage, which places the homepage title alongside the "site" title. App directory's Metadata API has this templating set up, but for the root page only shows homepage title)
        - Description
        - Site manifest
        - Locale
        - OpenGraph image
        - Twitter image, title, and description
        - favicons and apple touch icons
- Preview
    - Should preview with draft content
- Revalidation
    - Should revalidate. In Craft it may either show a failure, or no result. No result matches the behavior of the investigations. It's unclear why exactly that is since the app directory revalidation does return a `200` and `revalidation: true`. 
- Localization
    - Should show content in current locale
    - Should direct user to preferred locale
    - Switcher should switch between English and Spanish (hopefully faster than previously)
    - `lang` attribute on HTML root should change with locale
- Error page
    - Should go to error page if a page component throws an error. This is trickier to test, but React's dev-tools can force components into an errored state image
 - 404 page
     - Should go to 404 page if an invalid URI is entered (just type "Dfasdflasdkfja;sldfkjsdlf"). It should also attempt to show the page in the user's preferred locale

On other pages:

The most critical thing to check on other pages is just that nothing is obviously broken from moving the homepage. There are some minor changes that touch other pages, but their impact should be invisible:

- memoification of NoirLab release queries after they started endlessly looping on homepage
- moving some utility functions to a new module so they can run on the server vs. the utils file that had client-side imports
- moving a couple pieces of data from a context to props in the header
- wrapping a suspense boundary around the auth modals so they don't force pages into clientside rendering


hooks/useAuthentication.js
```diff
@@ -55,15 +55,8 @@ function hasPendingDeletion(userData) {
...
-function normalizeLanguage(language) {
-  if (language !== "en" && language !== "es") return null;
-  if (language === "en") return "en-US";
-  return language;
-}
-
 // TODO: store refresh token in cookie so token can be refreshed after browser session ends
-export default function useAuthentication(data) {
-  const { push } = useRouter();
+export default function useAuthentication() {
```
alexgoff, line 59: removed most of the redirecting to preferred language stuff, our routing handles this

lib/api/globals/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  siteInfoFragment,
+  footerFragment,
+  rootPageInfoFragment,
+  contactFormFragment,
+} from "@/lib/api/fragments/global";
+import { categoriesFragment } from "@/lib/api/fragments/categories";
+import { userProfileFragment } from "@/lib/api/fragments/page";
+import { getSiteFromLocale } from "@/lib/helpers/site";
+
+export async function getGlobalData(locale = fallbackLng) {
```
alexgoff, line 15: moved out to a new file, app directory porting will require moving some stuff around since a lot of API calls are done in files that are mixed with `useSWR` which is client-only.

lib/api/metadata/index.ts
```diff
@@ -0,0 +1,45 @@
+import { queryAPI } from "@/lib/fetch";
+import { getSiteFromLocale } from "@/lib/helpers/site";
+import { fallbackLng } from "@/lib/i18n/settings";
+import { gql } from "graphql-request";
+
+export async function getEntryMetadataByUri(
```
alexgoff, line 6: create a small fetch utility for getting the title/description for pages to use in metadata

lib/api/noirlabReleases.js
```diff
@@ -51,9 +53,10 @@ export function useRelease(site, entry) {
 
 export function useReleases(site, entries = []) {
   const locale = getLocaleString(site);
-  const releaseEntries = entries.filter((entry) => entry.pressReleaseId);
-
-  const getRel = (url) => fetch(url).then((res) => res.json());
+  const releaseEntries = useMemo(
```
alexgoff, line 56: this had to be memo-ified because it was re-rendering endlessly. Not sure what caused it or why it turned up now, chalked it up to React stuff that happens when you move project structures around

lib/helpers/site.ts
```diff
@@ -0,0 +1,7 @@
+export const getSiteFromLocale = (locale: string) => {
```
alexgoff, line 1: extracted these from the big utils file which mixed client and server code, separation of the two is critical for app directory

lib/i18n/server.ts
```diff
@@ -0,0 +1,11 @@
+import { cache } from "react";
+import { cookies, headers } from "next/headers";
+import { cookieName, fallbackLng } from "./settings";
+
+export const getLocale = cache(() => {
```
alexgoff, line 5: utility for getting the locale on the server outside of the `[locale]` directory context

lib/styles/font.ts
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line camelcase
+import { Source_Sans_3 } from "next/font/google";
+
+const SourceSansPro = Source_Sans_3({
+  weight: ["400", "700", "900"],
```
alexgoff, line 5: replace our font setup with next/font's cached Google fonts

lib/styles/registry.tsx
```diff
@@ -0,0 +1,29 @@
+"use client";
+
+import React, { useState } from "react";
+import { useServerInsertedHTML } from "next/navigation";
+import { ServerStyleSheet, StyleSheetManager } from "styled-components";
+
+export default function StyledComponentsRegistry({
```
alexgoff, line 7: styled components setup

pages/api/preview.js
```diff
@@ -52,7 +52,7 @@ const preview = async (request, response) => {
   response.setDraftMode({ enable: true });
 
   if (res.entry.uri === CRAFT_HOMEPAGE_URI) {
-    const redirect = `/${locale}`;
+    const redirect = `/${locale}?preview=${previewToken}`;
```
alexgoff, line 55: add the previewToken for app directory

pages/api/revalidate.js
```diff
@@ -31,11 +34,14 @@ async function handler(req, res) {
   }
 
   try {
-    const searchParams = new URLSearchParams(query);
-
-    await fetch(`${BASE_URL}/api/app-revalidate?${searchParams.toString()}`);
-    await res.revalidate(`/${query.uri}`);
-    return res.status(200).json({ revalidated: true });
+    if (APP_ROUTER_REVALIDATE.includes(query.uri)) {
```
alexgoff, line 37: try to minimize errors where revalidation occurs in both environments but one is missing data, this can be removed when migration is complete

theme/styles/base/_base.scss
```diff
@@ -3,16 +3,6 @@
 @use "abstracts/mixins/base";
 @use "abstracts/mixins/layout";
 
-// Base
```
alexgoff, line 6: CSS loading order got a little mixed up and these styles that were previously overwritten surfaced. The box-sizing is unnecessary since we already have a reset that handles it, and the `[loading]` is outdated and ends up hiding the images since most image elements have a loading attribute used to specific if they are lazy or eager

ericdrosas87 (approved):

PR #632, opened 2024-12-10
fix: dates inconsistent across site

alexgoff:
After the previous fix for dates to normalize timezones coming from Craft, some dates were inconsistent in different areas of the site. 

There were multiple date utilities with duplicate code, and unclear pattern of use. To resolve this, the number of datestring utilities has been condensed down to two: one to generate strings and another to generate objects of parts. 

Both utilities go through the same Craft date normalization by default.


helpers/dates.ts
```diff
@@ -0,0 +1,72 @@
...
+  date.setTime(date.getTime() + date.getTimezoneOffset() * 60 * 1000);
+};
+
+interface DateStringOptions {
+  locale?: string;
+  isShort?: boolean;
+  isCraftDate?: boolean;
+}
+
+export const makeDateString = (
```
alexgoff, line 13: combination of existing `makeDateString` and `useDateString`

PR #682, opened 2025-01-13
feat: galleries in search page

alexgoff:
Add galleries to the search results page.

Includes refactor to the page itself. Previously it was a template served through the `[..uriSegments]` catch-all. I think that did not make sense for a one-off page that would always live at the same route `/search`. I separated the search page into it's own static route so now the other pages don't bundle search code and search doesn't bundle code for other pages.


app/[locale]/[...uriSegments]/page.tsx
```diff
@@ -107,7 +106,6 @@ const sectionMap = {
   events: EventPageTemplate,
   glossaryTerms: GlossaryPageTemplate,
   news: NewsPageTemplate,
-  searchResults: SearchPageTemplate,
```
Comment (alexgoff) [inline app/[locale]/[...uriSegments]/page.tsx:110]: remove from catch-all

app/[locale]/search/page.tsx
Comment (alexgoff) [inline app/[locale]/search/page.tsx:1]: implements what was previously the `SearchPageTemplate` as a static route

components/dynamic/SearchList/index.js
```diff
@@ -101,10 +102,16 @@ const SearchList = ({
           {entries?.length > 0 && (
             <Grid columns={1}>
               {entries.map((entry) => {
+                const { cantoAssetSingle } = entry;
+
                 return entry.id ? (
                   <Tile
                     key={entry.id}
-                    image={entry.image?.[0]}
+                    image={
```
alexgoff, line 110: show Canto assets in search results if available

lib/api/entries/index.ts
```diff
@@ -6,45 +6,24 @@ export async function getEntriesByLocale(locale: string) {
...
-      query getEntriesBySite($site: [String]) {
-        entries(
-          site: $site
-          section: ["pages", "searchResults"]
-          type: ["not", "redirectPage"]
-        ) {
-          uri
-        }
+  const query = gql`
+    query getEntriesBySite($site: [String], $level: Int) {
```
alexgoff, line 10: since search pages aren't part of the catchall route anymore, they don't need to be part of it's static params generation

ericdrosas87 (approved):

PR #689, opened 2025-01-17
feat: main gallery page

alexgoff:
Adds a main gallery to the `/gallery` route. Should be used in conjunction with `ag-677` API branch.

- Delete existing gallery content from Craft pages structure
- Add a new /gallery and set page type to `Gallery Landing Page`
- Assign a gallery
- Go to /gallery and see your gallery
- Navigate to assets in the gallery
- Navigate to /gallery/[your-main-gallery's-slug] and see that it redirects back to /gallery


components/organisms/Header/Subnavigation.js
```diff
@@ -39,9 +39,11 @@ export default function Subnavigation({
         const hasChildren = children && children.length > 0;
         const isActiveSub = id === activeSub;
 
+        if (!uri && !hasChildren) return null;
```
alexgoff, line 42: hide children that do not have a URI (in this case: Collections)

lib/api/globals/navigation.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (!data || !data.navigationItems) return [];
+
+  const { navigationItems, galleriesEntries } = data;
+
+  const { data: structure = [] } =
+    navigationStructure.safeParse(navigationItems);
+
+  const galleryIndex = structure.findIndex((item) => item.uri === "gallery");
+
+  if (galleryIndex > -1) {
```
alexgoff, line 80: put the navigation through some parsing, then stuff in galleries into their custom spot

app/[locale]/[...uriSegments]/layout.tsx
Comment (alexgoff) [inline app/[locale]/[...uriSegments]/layout.tsx:1]: added a `metadataMap` so that pages generated by `uriSegments` can figure out their own metadata instead of putting it into giant queries here. Separated out into the layout alongside static param generation for organization

app/[locale]/[...uriSegments]/page.tsx
```diff
@@ -112,9 +22,13 @@ const sectionMap = {
   staffProfiles: StaffPageTemplate,
 };
 
+const pageMap = {
```
Comment (alexgoff) [inline app/[locale]/[...uriSegments]/page.tsx:25]: we didn't have a way of mapping sub-types of page entries before

ericdrosas87 (approved):

PR #709, opened 2025-02-10
refactor: preview token into session

### LINKED ISSUE #708: fix: inconsistent preview display
Previews are not consistent across pages in the application.

Currently, the preview token for querying Craft drafts is send via a URL parameter, which pages have to extract and pass to the GraphQL client on each query.

Refactor to set a session cookie from the `/preview` endpoint containing the preview token. In the GQL client, read the preview token from the cookies if Next's draft mode is enabled and attach to the query.

alexgoff:
Resolves #708 

Instead of manually affixing the Craft preview token to the search params, the token is committed to a session cookie which is automatically appended by the GraphQL client on queries. 

Since both Next draftMode and the preview token cookies are session cookies, they are not cleared until the browser is closed. To avoid getting stuck in preview mode, a banner is shown indicating preview mode with a button to disable it.

Testing
- Check out a live preview in Craft like a page, gallery, or news article
- Check out the non-live preview ("View") and see that the preview mode banner appears
- Click "disable" on the preview mode banner and see that any draft changes disappear

Does the language on the banner make sense? Would "Exit" be a better CTA on the button?

### DIFF
--- a/components/organisms/PreviewMode/Banner/index.tsx
+++ b/components/organisms/PreviewMode/Banner/index.tsx
@@ -1,0 +1,43 @@
+"use client";
+import { FC, useEffect, useState } from "react";
+import { useRouter } from "next/navigation";
+import { useTranslation } from "react-i18next";
+import IconComposer from "@rubin-epo/epo-react-lib/IconComposer";
+import endPreviewMode from "@/services/actions/endPreviewMode";
+import styles from "./styles.module.css";
+
+const Banner: FC = () => {
+  const { t } = useTranslation();
+  const [isLivePreview, setLivePreview] = useState(false);
+  const { refresh } = useRouter();
+
+  useEffect(() => {
+    if (window && window.top !== window.self) {
+      setLivePreview(true);
+    }
+  }, []);
+
+  const handleDisable = async () => {
+    await endPreviewMode();
+
+    refresh();
+  };
+
+  if (isLivePreview) return null;
+
+  return (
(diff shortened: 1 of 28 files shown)


components/organisms/PreviewMode/Banner/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const Banner: FC<BannerProps> = ({ isEnabled }) => {
+  const { t } = useTranslation();
+  const [isLivePreview, setLivePreview] = useState(false);
+  const { refresh } = useRouter();
+
+  useEffect(() => {
+    if (window && window.top !== window.self) {
+      setLivePreview(true);
```
alexgoff, line 20: checks if currently inside an iframe and does not display if so

services/sessions/preview.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const previewSession = () => {
+  const cookieSettings: Partial<ResponseCookie> = {
+    httpOnly: true,
+    sameSite: true,
+    secure: true,
+  };
+  const tokenCookie = "previewToken";
+
+  const draft = () => {
```
alexgoff, line 12: wrapping this inside of try-catch because if the client is used for `generateStaticParams` then draft mode is not available and throws an error. 

Alternatively a separate client could be created that doesn't include the preview mode checks, but quashing the error seems fine to me.

ericdrosas87 (approved):

PR #744, opened 2025-03-19
build: add T3 for type-safe environment

### LINKED ISSUE #712: build: set up type-safe environment variables
So that deployments never end up in a broken state do to misconfigured environment, and that developers can have a strongly typed interface with the Next environment variables, set up [T3](https://github.com/t3-oss/t3-env) and write configuration for the current environment variables. Replace all existing uses with the new T3 import

https://mwskwong.com/blog/why-t3-env-is-my-go-to-for-managing-environment-variables

alexgoff:
Resolves #712 

Installs [T3](https://env.t3.gg/) into the rubin-obs-client with a type-safe config defined and most environment variable accesses updated from `process.env -> env.ts` 

A type-safe environment defines a schema which can be used to determine if a build is valid or not. The schema is imported into the `next.config.js` on build, and if the environment is invalid the build will error and not move forward. 

With this in place, all builds inherently have a working environment and it becomes impossible for missing env vars to cause issues in a deployment. An additional benefit is that we can set defaults, refine types, and have typed env vars in the client code. This can remove logical fallbacks like setting empty strings if an env var containing a URL is not present, for example.

**Testing**

- Run `yarn` to pick up the new packages
- Start a production build with `yarn bs`
- There is a very good chance your existing `.env` file is missing some values considered essential. Update with these values and rebuild.
- If it built successfully, try removing a variable from `.env` and re-build to see the error.
- Non-Next stuff that uses the .env
  - Cypress: `yarn test`
  - GraphQL codegen: `yarn codegen:gql`
  - NOIRLab codegen: `yarn codegen:noirlab`


env.config.ts
alexgoff, line 1: loader for the Next `.env` when used outside of Next (Cypress, codegen)

env.ts
```diff
@@ -0,0 +1,46 @@
+import { createEnv } from "@t3-oss/env-nextjs";
+import { z } from "zod";
+
+export const env = createEnv({
```
alexgoff, line 4: defines our environment explicitly, with schema validation

middleware.ts
alexgoff, line 1: building on the changes from earlier this week, `middleware` is now converted to Typescript and uses T3 in all cases except `process.env.HOSTNAME` which is not defined at buildtime(?)

next.config.js
```diff
@@ -1,38 +1,19 @@
 import { join, dirname } from "path";
 import { fileURLToPath } from "url";
+import { createJiti } from "jiti";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
-let API_URL;
```
alexgoff, line 7: `jiti` allows TypeScript files to be imported into next.config, could be removed after updating to Next 15.

This import will type-check the environment at build time. Any invalid env will error the build process immediately.

ericdrosas87 (approved):

PR #756, opened 2025-03-27
feat: content block for Skyviewer

### LINKED ISSUE #690: feat: Skyviewer content block
Set up a new content block for Skyviewer entries that renders the provided URL into an iframe. Use a stored env variable for the hostname, and select explorer or embed based on the configured option from Craft.

For all blocks, the iframe needs the attributes:
- [allow-downloads](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#allow-downloads)
- [allow-scripts](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#allow-scripts)

For embeds add attributes:
- [allow-popups-to-escape-sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#allow-popups-to-escape-sandbox)

Explorer blocks should be styled to span the full width of the screen, while embeds should be restricted to a container

alexgoff:
Resolves #690 

- adds a Skyviewer content block
  - no tests unfortunately since Cypress does not support server components in unit tests
  - move fragment to type-checked services folder
  - handles full width and embedded versions of the Skyviewer
- adds RichTextContent atom with tests
- extends server/client boundary to cover put `ContentBlockFactory` on the server on every page it is used so that server-only components like the Skyviewer can render

**Testing**

- Use in tandem with the [API branch](https://github.com/lsst-epo/rubin-obs-api/pull/355)
- add `SKYVIEWER_BASE_URL` to your `.env.local`
- On a new or existing page, add a Skyviewer content block (the homepage is fine)
- Configure with RA/DEC/FOV and other values (you can get these values by visiting the dev instance of the Skyviewer, and getting the URL from the share menu, it'll be in the URL params)
- Visit page
  - See that your content block is visible
  - See that Skyviewer functionality is present, particularly share features that can be blocked by iframe permissions
  - Make changes, check out the full width vs. embed, see that the changes are reflected
- Repeat by adding Skyviewer content block to a non-press release news article

**TODO**

Need to create a followup ticket to work with the content team on handling misconfigured Skyviewers, or Skyviewers that have an error when their iframe loads.

### DIFF
--- a/services/craft/entries/content-blocks/skyviewer.ts
+++ b/services/craft/entries/content-blocks/skyviewer.ts
@@ -1,0 +1,31 @@
+import { graphql } from "@/gql";
+
+const SkyviewerBlockFragment = graphql(`
+  fragment skyviewerBlock on contentBlocks_skyviewer_BlockType {
+    id
+    embedTitle
+    typeHandle
+    captionRichText
+    dec
+    fov
+    ra
+    fullWidth
+    backgroundColor
+  }
+`);
+
+export const SkyviewerNewsBlockFragment = graphql(`
+  fragment skyviewerNewsBlock on contentBlocksNews_skyviewer_BlockType {
+    id
+    embedTitle
+    typeHandle
+    captionRichText
+    dec
+    fov
+    ra
+    fullWidth
+    backgroundColor
+  }
+`);
+
+export default SkyviewerBlockFragment;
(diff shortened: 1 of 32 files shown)


services/craft/entries/content-blocks/skyviewer.ts
alexgoff, line 1: regular content blocks and news content blocks are technically different queries with different type signatures, even if they query the same data

PR #788, opened 2025-04-04
perf: replace classnames with clsx

### LINKED ISSUE #726: perf: replace classnames with clsx
**Context**

[classnames](https://www.npmjs.com/package/classnames) is a widely used package to concatenate CSS class names in React components using arrays and objects. While the package is extremely popular, some newer, smaller, and faster alternatives now exist, particularly clsx which has a much smaller bundle size and is about[ 2-3x faster than classnames](https://github.com/lukeed/clsx/tree/925494cf31bcd97d3337aacd34e659e80cae7fe2/bench).

Since most of our use of classnames happens during server renders, a more performant package will reduce server load and improve render times. 

**Task:**

Replace the classnames package with more performant [clsx](https://www.npmjs.com/package/clsx)

clsx is a 1:1 drop-in replacement for classnames, classnames can be removed at the completion of this task

soshwosh:
Closes #726

### DIFF
--- a/components/atomic/Button/Hamburger/index.tsx
+++ b/components/atomic/Button/Hamburger/index.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { FunctionComponent } from "react";
-import classNames from "classnames";
+import clsx from "clsx";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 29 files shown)


alexgoff (approved):
First one down!

PR #796, opened 2025-04-09
feat: alternate share content block

### LINKED ISSUE #776: feat: share content block redux
Changes to the existing "share" content block:

- Selection between existing "small" and new "callout" sized block
- Callout block with
  - Background color selection
  - Title
  - description
  - Mobile size uses web share API
- Update social shares to include:
  - Facebook
  - X (Twitter)
  - LinkedIn
  - Email
  - Copy URL

Image

alexgoff:
Resolves #776 

Adds an alternate display of the existing share content block with:
- updated icon buttons for all variants
- new larger variant, with callout-style backgrounds
- new field for entering text in Craft
- mobile display using `SharePopup`
- native Web Share API where available, when using `SharePopup`

Test:
- Use with [API branch](https://github.com/lsst-epo/rubin-obs-api/pull/360)
- Find an existing content block that uses Share, or add a new Share content block
- First set up a "small" variant
  - See that Facebook, X (Twitter), LinkedIn, email, and copy buttons display
  - See that at mobile sizes, content is rendered in share popup
- Set up a "large" variant (reference issue for intended display)
  - Set some text
  - Set a background color
  - See that your text is applied
  - See that your background color is applied
  - See that at mobile sizes, a large share button is shown that launches the share popup
  - If running in the devtools responsive mode, as a mobile device like an iPhone or Pixel device, clicking the share button should launch native share features

Component tests added for SharePopup, ShareBlock uses server features and can't be tested in a Cypress component test


components/molecules/SharePopup/index.tsx
```diff
@@ -39,19 +50,50 @@ const SharePopup: FC<SharePopupProps> = ({ title, url, className }) => {
     }
   );
 
+  const finalUrl = createFinalUrl(url || pathname || undefined);
+
+  const handleClick: MouseEventHandler = async (event) => {
```
alexgoff, line 55: click handler will block the default event, to prevent the popover from showing, if the device's user agent matches for mobile and has the sharing API available

components/content-blocks/ShareBlock/index.tsx
alexgoff, line 1: converted to TypeScript, uses typed fragment, adds new variant

components/content-blocks/ShareBlock/styles.module.css
alexgoff, line 1: phablet is our target breakpoint, iPads want to be treated like desktop devices so they get regular share buttons

components/molecules/SharePopup/SharePopup.cy.tsx
alexgoff, line 1: kind of a rudimentary test but many features like user agent strings can't be run in Cypress component tests or even e2e tests

PR #807, opened 2025-04-21
feat: dark mode for text, image, video, and related content

alexgoff:
Adds dark mode styles to select content blocks. A less fragmented approach would be to enable dark mode at the page level, however the site is not well set up for this at the moment and many pieces of UI have unique styles that are geared for light more and would not display well with a blanket global dark mode. 

This change aims instead of specifically target the content blocks needed for the first look minisite and add background color options for those blocks.

Use with API branch.

Test:
- Go to an existing content block for text, images, videos, or related content (or create a new one)
- Look for the new "Background Color" tab at the top of the block
- Set the background color
- On full-screen pages (that do not have "Overlap hero" selected), the background color should span the width of the page content. On hero overlap pages it will only span the article width
- If "black" is selected as a color, these content blocks should display in a dark mode variant


components/content-blocks/GridBlock/index.js
alexgoff, line 1: little bit of CSS-in-JS conversion at the same time

components/atomic/Tile/styles.js
alexgoff, line 1: do some basic style updates on tiles, trying to touch as little as possible while still getting done what's necessary since this is a pretty complex component.

components/content-blocks/Image/index.tsx
alexgoff, line 1: have to re-write some of the `Image` content block since it has floats built in and needs updated behavior to work well with figure backgrounds in dark mode

components/content-blocks/Video/index.tsx
alexgoff, line 1: video is mostly unaltered apart from adding the dark mode background except that the YouTube loader is lazy loaded with a skeleton stand-in

services/craft/entries/content-blocks/text.ts
alexgoff, line 1: strongly type these fragments

PR #818, opened 2025-04-29
Fixed pagination formatting issues

soshwosh:
Added conditional formatting to fix the errors with rendering page numbers and ellipses in the pagination component.
Added Cypress component tests to verify changes.


components/molecules/Pagination/Pagination.cy.tsx
```diff
@@ -0,0 +1,270 @@
...
+        "3",
+        "4",
+        "5",
+        "6",
+        "7",
+        "8",
+        "...",
+        "11",
+      ]);
+      cy.get(':nth-child(3) > [rel="prev"]').should("not.exist");
```
alexgoff, line 166: suggestion: the next/prev buttons don't need to be part of these tests, their coverage can just be contained to tests that do a check for their existence and values based on inputs. A change to next/prev could end up breaking a bunch of unrelated tests and making it harder to isolate what happened.

components/molecules/Pagination/Pagination.cy.tsx
```diff
@@ -0,0 +1,270 @@
...
+        </I18NextClientProvider>
+      );
+
+      cy.get("ul > li").should("have.length", 10);
+      cy.get("ul > li").each(($element) =>
+        paginationElements.push($element.text())
+      );
+      cy.wrap(paginationElements).should("deep.equal", [
+        "1",
+        "...",
```
alexgoff, line 200: suggestion: for these ellipses tests, this is very comprehensive but it's also testing beyond the core of what this test is about which is: is there an ellipses at index 1 and index length - 1 under the conditions we expect?

The rest of the numbers could be part of any range in the pagination but they don't really matter for this test.

components/molecules/Pagination/Pagination.cy.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+    mount(
+      <I18NextClientProvider locale="en">
+        <SearchParamsContext.Provider value={searchParams}>
+          <Pagination limit={limit} offset={offset} page={page} total={total} />
+        </SearchParamsContext.Provider>
+      </I18NextClientProvider>
+    );
+
+    cy.viewport(1440, 900);
```
alexgoff, line 22: suggestion: if you're using a viewport repeatedly in tests, it can be defined in the test config: https://docs.cypress.io/api/commands/viewport#Set-viewport-in-the-test-configuration

alexgoff (changes requested):
Paginator fixes are working great! Just some stray logs and comments. I left some notes on the unit tests for how to re-use some of your testing boilerplate and simplify the tests themselves.

Really comprehensive work!

components/molecules/Pagination/Pagination.cy.tsx
soshwosh: Since the pagination component should have max 10 elements, I hardcoded the check for the ellipses to indexes 1 and 8 of the element list. If you think that max of 10 is subject to change, or that it's just better practice, I can programmatically check the index based on the length of the element list - 1.

PR #819, opened 2025-04-29
feat: full screen video blocks

alexgoff:
Adds video blocks with Canto sources, and an option to display video blocks in a dialog above all other content (for first look intro video)

Testing:
- use with API branch [am-775](https://github.com/lsst-epo/rubin-obs-api/pull/378)
- in the homepage entry, create a new video block
  - select Canto as source typpe
  - pick one or two sources from Canto (M1M3 Install has both landscape and portrait orientations)
- on the "style" tab of the block, turn fullscreen on
- visit the homepage, expect to see your video fade in after loading
- click the close button, hit escape, or wait for the video to finish
- refresh page, expect to not see video again (value is in local storage if you want to delete it)

Component tests added for new `CantoPlayer` for video blocks


components/atomic/Button/IconButton/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      visibleText,
+      isBlock = false,
+      icon,
+      size,
+      onClickCallback,
+      buttonProps,
+      iconProps,
+      className,
+    },
+    ref
```
alexgoff, line 34: the only change here, apart from moving into TypeScript, is adding a `ref` to the button

package.json
```diff
@@ -110,6 +110,7 @@
     "tz-lookup": "^6.1.25",
     "urql": "^4.1.0",
     "use-resize-observer": "^9.0.2",
+    "usehooks-ts": "^3.1.1",
```
alexgoff, line 113: `usehooks-ts` is a handy dandy utility library that we've been using in other projects, here it's brought in for it's React local storage interface but it has many other utilities we can make use of

PR #820, opened 2025-05-02
refactor: static rendered homepage

alexgoff:
**Experimental**

Re-work to enable static rendering. For server components to have access to the `locale` to figure out the language to render in, they previously needed to either have the `locale` prop drilled deeply into the render tree, or to read it off the request headers which forced dynamic rendering.

`next-intl` has improved this situation by caching the `locale` param at the page level, so it can be used by any child component of the page request, re-enabling static rendering for a route.

A basic `next-intl` setup has been implemented, replacing our previous `getLocale()` method that server components used. This is sort of a wrapper for the actual internationalization tool that we use, `i18next`, which is a dependency of the libraries. 

The homepage has been put into static rendering, while all others are forced dynamic to preserve their existing behavior. These could be later put onto static rendering.

**Old setup**
```jsx
<Page /> <-- has locale from route segment
   ...
       <DeeplyNestedComponent /> <-- locale either needs to be drilled down to here, or getLocale will read from headers and become dynamic
```

**New setup**
```jsx
<Page /> <-- has locale from route segment, caches it with setRequestLocale
   ...
       <DeeplyNestedComponent /> <-- getLocale retrieves from request cache if set, otherwise will read from headers and become dynamic
```


components/organisms/Header/LanguageSelect/index.js
```diff
@@ -39,18 +28,8 @@ export default function LanguageSelect({ className }) {
...
-      if (locale !== fallbackLng) {
-        parts.shift();
-      }
-
-      const route = `/${newLocale}/${parts.join("/")}${filterSearchParams(
-        searchParams
-      )}`;
       setLoading(true);
-      router.replace(route, { scroll: false });
+      router.replace({ pathname, query: searchParams }, { locale: newLocale });
```
alexgoff, line 32: use `next-intl` router helper that will change the locale cookie on locale change

PR #837, opened 2025-05-15
fix: set revalidation on NOIRLab requests

### LINKED ISSUE #836: fix: route segment revalidation not working
The route segment revalidation for NOIRLab API pages is not sticking, all revalidation periods for fetches are fixed at `31536000`

Move revalidation period from route segment level to individual fetches

alexgoff:
Resolves #836 

Next route segment revalidation behavior does not work as described in the docs, so to ensure we cache our NOIRLab content at the correct intervals, add an interceptor to the NOIRLab client that configures these settings on each request.

- Generated content has been moved from /services/noirlab to /services/noirlab/schema
- The API has been re-exported from /services/noirlab with interceptors applied
- A request interceptor has been added that sets a revalidation time using an environment variable and forces caching

**Testing**
- If you want to test the existing behavior on `develop`
  - Start dev server
  - Clear `.next/cache/fetch-cache`
  - Visit a page like `gallery/collections/news-images/noirlab2425a`
  - Look in fetch-cache for an entry that goes to the NOIRLab endpoint (there will be 4-5 caches from this first page load)
  - See that the revalidation time is 31536000
- To test the fix, repeat this behavior after clearing the cache again and see that the revalidation time is the default 300 seconds (5 minutes)
- Visit a press release, see that the revalidation time is 31536000 since this caching is only applied to NOIRLab media, not news content that we update through on-demand revalidation


services/noirlab/index.ts
alexgoff, line 1: all request interceptors are imported here and applied, the SDK and types are then re-exported. Classes from the SDK will use this modified client with interceptors applied

env.ts
```diff
@@ -17,6 +17,10 @@ export const env = createEnv({
     CRAFT_SECRET_TOKEN: z.string().min(1),
     GOOGLE_APP_SECRET: z.string().min(1),
     SKYVIEWER_BASE_URL: z.string().url(),
+    NOIRLAB_REVALIDATE: z.coerce
```
alexgoff, line 20: add an env var for revalidation period so we can alter it without a code change

services/noirlab/interceptors/request.ts
```diff
@@ -0,0 +1,23 @@
+import { RequestOptions } from "@hey-api/client-next";
+import { env } from "@/env";
+
+const isServer = typeof window === "undefined";
+
+const addCaching = (options: RequestOptions) => {
+  const isMedia = new RegExp(/\/media\/(images|videos)\//).test(options.url);
+
+  if (isServer) {
+    if (!options.cache) {
```
alexgoff, line 10: respect any cache already applied, otherwise explicitly force cache. This will be important for a Next 15 upgrade where caching is no longer the default.

services/noirlab/interceptors/request.ts
```diff
@@ -0,0 +1,23 @@
...
+      options.cache = "force-cache";
+    }
+
+    if (isMedia) {
+      options.next = {
+        revalidate: env.NOIRLAB_REVALIDATE,
```
alexgoff, line 16: set our revalidate but respect any overrides

ericdrosas87 (approved):
Ah okay, this makes sense

PR #844, opened 2025-05-16
feat: add individual image pages to sitemap.xml

soshwosh:


lib/api/sitemap/index.ts
```diff
@@ -78,6 +78,39 @@ export const getSiteMapNewsData = async (locale: string) => {
...
 
+export const getSitemapGalleryData = async (locale: string) => {
+  const site = getSiteFromLocale(locale);
+
+  const query = gql`
+    query ImageSitemapData($site: [String]) {
+      galleries: galleriesEntries(site: $site) {
+        ... on galleries_gallery_Entry {
+          uri
+          assetAlbum {
```
alexgoff, line 89: issue: Canto albums can contain up to 6 different asset types: image, video, document, audio, presentation, and other. 

Add a filter here so that we only query supported schemes. Check `lib/api/galleries/schema.ts` for the `SupportedCantoSchemes` schema and add them with a `whereIn` argument like: `whereIn: {key: "scheme", values: ["image","video","document"]}`

check out `getGalleryData` in `lib/api/galleries/index.ts` for an example of how the bindings are set up

lib/api/sitemap/index.ts
```diff
@@ -78,6 +78,39 @@ export const getSiteMapNewsData = async (locale: string) => {
   return { siteTitle, news: newsEntries };
 };
 
+export const getSitemapGalleryData = async (locale: string) => {
```
alexgoff, line 81: suggestion: GraphQL type generation isn't enabled for this file, but it could be to give us deeper insight into the return

- in `codegen.ts` add `"lib/api/sitemap/*.ts"` to the array of documents
- in your query, replace `gql` with `graphql()` which you can import from `import { graphql } from "@/gql/gql";`
- a type-checking error will appear, run `yarn codegen:gql` and this will generate a schema for your query
- the canto assets will likely be untyped because there's some issue with the plugin

app/[locale]/sitemap.xml/route.ts
```diff
@@ -47,6 +48,24 @@ export async function GET(
     return entry;
   });
 
+  const { galleries } = await getSitemapGalleryData(locale);
+  const imageData: any[] = [];
+  galleries.forEach((gallery) => {
+    const { uri, assetAlbum } = gallery;
+
+    assetAlbum.forEach((asset) => {
```
Comment (alexgoff) [inline app/[locale]/sitemap.xml/route.ts:56]: issue: this setup is removing all gallery pages (including landing pages) except for image asset pages.

You could map `galleries` to generate the sitemap entries for the gallery landing page, and then maintain a separate array that holds the assets, and then just use the filetype to optionally append the `image:image` addon

alexgoff (changes requested):

lib/api/sitemap/index.ts
soshwosh: I started with this but haven't resolved all the type errors yet. Still using `gql` for now

soshwosh: Completed adding this in this commit - https://github.com/lsst-epo/rubin-obs-client/pull/844/commits/1ec3ee059ccd158ba917991f0d2d34ab9e704be1

alexgoff (approved):

PR #900, opened 2025-06-10
refactor: update an older component from styled-components to CSS modules

### LINKED ISSUE #880: refactor: update an older component from styled-components to CSS modules
**Context**

Many of our older components use [styled-components](https://styled-components.com/) for their styling. styled-components is a runtime CSS-in-JS solution that is very versatile, but is reaching the end of its life. 

There are several limitations to styled-components that make them less desirable than they once were:

- can't be rendered fully on the server because they use context
  - final render must be done through a client component
- can be rendered as a first-pass serverside render, but still require client hydration

The solution going forward is [CSS modules](https://github.com/css-modules/css-modules) which is simply a build-time JavaScript interface on top of CSS, to solve scoping. 

CSS modules can be rendered at build time and CSS does not require any hydration apart from the normal paint that the browser must do to render the page. They are not reliant on a third party's support, so they can't reach an end-of-life. If the module spec ever becomes unsupported, the files simply can become global CSS. 

For server rendering, they can help [reduce RSC payload](https://vercel.com/guides/how-to-optimize-rsc-payload-size#client-component-placeholders) by reducing the number of unique client components that are rendered and begin to chip away at large client trees

Minor post-processing enhancement is handled by PostCSS. This is less brittle than SCSS since there are no built-in behaviors like mixins or functions that rely on the SCSS compiler, CSS with PostCSS enhancement is almost always valid CSS even without the post-processor.

**Task**

Update the `Hero` molecule component to use CSS modules instead of styled-components.

- create a new styles.module.css file
- create classes to replicate each styled component declared in styles.ts
- copy over all styles
  - styles.ts uses a JavaScript helper called fluidScale to generate a clamp for fluid sizing of the height of the hero image between screen sizes. This can be replaced with a PostCSS processor for [utopia.clamp](https://utopia.fyi/blog/postcss-utopia/). Utopia clamp assumes pixel values, so pass in only integers to the function, ex. `utopia.clamp(100, 200)` vs. `fluidScale("100px", "200px")`
- import the CSS modules into the hero, replace each `<Styled.*` with their equivalent HTML element, and add the CSS module class.
- for the root element of the component, concatenating the `className` prop onto the `className` attribute will be necessary with `clsx`.
  - in almost all cases, the order should be base styles first, followed by the `className` prop to avoid [CSS load ordering issues](https://dev.to/receter/avoiding-css-order-of-appearance-problems-with-css-modules-3j7p)
- remove `styles.ts`

Verify visually

soshwosh:
Resolves #880

### DIFF
--- a/components/molecules/Hero/styles.module.css
+++ b/components/molecules/Hero/styles.module.css
@@ -1,5 +1,5 @@
 .heroContainer {
-  --Hero-height-fallback: utopia.clamp(540, 400);
+  --Hero-height-fallback: utopia.clamp(300, 400);
   position: relative;
   width: 100%;
   max-width: var(--size-width-center, 2000px);


yarn.lock
alexgoff, line 1: issue: since there's no package.json changes in this, remove the `yarn.lock` so we aren't getting side effect updates from `yarn upgrade`

alexgoff (changes requested):

PR #937, opened 2025-07-01
refactor: staticize entry callout

### LINKED ISSUE #922: refactor: statically render CalloutEntry block
Statically render the CalloutEntry block, so that it does not need to call the NOIRLab API on each refresh of the page.

alexgoff:
Resolves #922 

This is pretty close to completion, but the goal is to make the news and event callouts static renderable. The big barrier is that the news entries call the NOIRLab endpoint, and were doing so with a client-side `useRelease` hook.

That has been refactored into a static renderable method making use of the NOIRLabServices fetch methods.

The last part to do is just reassemble the pieces of these content blocks, and move styles over from styled-components to CSS modules.  NOIRLab will probably appreciate our users not hitting their API endpoint on each homepage load, and there won't be an image between the default site image and the image from NOIRLab when it loads.


components/content-blocks/Callout/CalloutNews/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const calloutImage = noirlabContent?.images
+    ? makeReleaseFeature(noirlabContent.images, "screen640")?.[0]
+    : (image?.[0] as ImageShape) || (hero?.[0] as ImageShape);
+
+  const darkMode = isDarkMode(backgroundColor);
+
+  const subtitle = entry?.description || noirlabContent?.subtitle;
+
+  return (
+    <Styled.Section
```
alexgoff, line 81: a bunch of these types need a little coercion, styled-components need their render properties like `$width` and `$overlay` explicitly declared, or the styled-component replaced with CSS modules

Issue #947, opened 2025-07-10
[B] Fix educators investigations landing page


ericdrosas87:
## Expected Behavior
The [educator investigations landing page](http://localhost:3000/education/educators/investigations/coloring-the-universe) should load as expected.

## Actual Behavior
The educator investigations landing page shows an error message about an undefined and missing suffix for alt text.

## Steps to Reproduce the Problem

  1. Navigate to the [educator investigations landing page](http://localhost:3000/education/educators/investigations/coloring-the-universe) 
  2. See the error

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #948, opened 2025-07-10
fix: added fallback logic to metadata->additional

ericdrosas87:
fix: added fallback logic to metadata->additional


soshwosh (approved):

ericdrosas87:
Resolves #947

PR #958, opened 2025-08-11
feat: remove contact form

ericdrosas87:
feat: remove contact form

### DIFF
--- a/components/organisms/Footer/index.js
+++ b/components/organisms/Footer/index.js
@@ -1,14 +1,14 @@
 import Social from "./Social";
 import Links from "./Links";
-import ContactForm from "./ContactForm";
+// import ContactForm from "./ContactForm";
 import { Image } from "@rubin-epo/epo-react-lib";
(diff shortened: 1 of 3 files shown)


ericdrosas87:
To test:

1. Clone this branch down
2. Run `yarn dev`
3. Load `http://localhost:3000` in the browser and scroll to the bottom of the page and see that the contact form is removed
4. Resize the browser window and confirm that the aspect ratios never look wonky

components/organisms/Footer/index.js
```diff
@@ -1,14 +1,14 @@
 import Social from "./Social";
 import Links from "./Links";
-import ContactForm from "./ContactForm";
+// import ContactForm from "./ContactForm";
```
soshwosh, line 3: (A clarification question, not a challenging one) I see the JSX is fully deleted but the imports are still here, just commented out. Are we leaving the commented-out imports because we think we'll need them again soon?

Issue #977, opened 2025-11-03
[F] Add Rubin Basics post page and feed


ericdrosas87:
Using the news post page as a starting place, I need to create a new page structure that shows a Rubin Basics Post. The publish date must not be visible on the page.

PR #1011, opened 2025-12-15
feat: added rubin basics page, feed, nav, etc

ericdrosas87:
This update has an accompanying API change already merged into `develop` of the `rubin-obs-api` repo.

This adds a new page to the navigation menu at `Explore`->`Rubin Basics`. This page acts as a feed for Rubin Basics - a new type of content closely modeled after the News entry type. There is tagged search, and Rubin Basics appear in search.


ericdrosas87:
This resolves #977

PR #1033, opened 2026-01-12
Er add survey progress widget

ericdrosas87:
To test:

1. Enable `Survey Progress` setting in `Summit Status Compact View` content block
2. Start up Next.js dev server
3. Assess `SurveyProgress` widget looks as expected in desktop view
4. Assess `SurveyProgress` widget looks as expected in mobile view

### DIFF
--- a/components/dynamic/SummitData/SurveyProgress/index.tsx
+++ b/components/dynamic/SummitData/SurveyProgress/index.tsx
@@ -1,0 +1,51 @@
+import { useSummitData } from "@/contexts/SummitData";
+import { useTranslation } from "react-i18next";
+import Loader from "@/components/atomic/Loader";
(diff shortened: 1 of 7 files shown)


components/dynamic/SummitData/SurveyProgress/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <WidgetSection
+      isCollapsible={false}
+      title={t("summit_dashboard.sections.survey_progress.title")}
+      caption={t("summit_dashboard.sections.survey_progress.caption")}
+    >
+      <div
+        className={clsx(styles.widgetBackground, styles.condensedBackground)}
+      >
+        {stillLoading ? (
```
ericdrosas87: This is the pattern we should use for displaying the loader in the `WidgetSection`

PR #1069, opened 2026-03-17
fix: switched widget grid display attribute from inline-grid to grid

### LINKED ISSUE #1061: Fix SSD exceeding website width on ultra-wide screen dimensions
Fix the Summit Status Dashboard so that it's constrained by the width of the rest of the webpage content on ultra-wide screens.

Image

soshwosh:
Resolves #1061 

Testing:
- Enable all 6 widgets in Craft. Check mobile, tablet, desktop, and ultrawide dimensions for:
- Outer margins around the entirety of the layout are present on desktop and ultrawide. No background for smaller dimensions.
- Widget width is less than height (widget is a narrow rectangle and not a square).
- Widget title, contents, caption and tooltip are centered vertically.
- All tooltips are centered above info icon.
- Each widget's contents are roughly the same height as one another.

Repeat the above with 5, 4, and 3 widgets enabled. You can go all the way down to 2 or 1 if you'd like, but we should never be showing the SSD with less than 3 widgets.

### DIFF
--- a/components/content-blocks/SummitStatusCompactView/styles.ts
+++ b/components/content-blocks/SummitStatusCompactView/styles.ts
@@ -2,7 +2,7 @@ import styled from "styled-components";
 import WidgetBackground from "@/components/atomic/WidgetBackground";
 
 export const Container = styled.div`
-  padding: 25px 0px 0px;
+  padding: 25px 0px;
   background-color: var(--neutral95, #1f2121);
 `;
 
--- a/components/layout/SummitStatus/WidgetContainer/styles.js
+++ b/components/layout/SummitStatus/WidgetContainer/styles.js
@@ -2,6 +2,10 @@ import styled from "styled-components";
 import { token } from "@/styles/globalStyles";
 
 export const WidgetContainer = styled.div`
+  display: flex;
+  flex-direction: column; 
+  align-items: center;
+  justify-content: center;
   padding: 15px 20px;
   margin-bottom: 1rem;
   color: #fff;


soshwosh:
@ericdrosas87 @jeffinnes A lot of these SVG overflowing, content de-centering issues seem to be arising from the browser struggling to reconcile clashes between the parent (`WidgetSection` and higher) rules and some implicit SVG and grid layout effects. A lot of "obvious" fixes (enforcing `align-items: center`, responsive `width`/`height`/`max-width`/`max-height`, auto margins, explicitly adding attributes that should've been implicitly applied like `preserveAspectRatio` on the SVGs, etc) were not applying as expected and causing more styling bugs. The grid layout was making it difficult to keep track of, and **ultimately I tried switching `SectionContent` to a flex display**. We aren't making use of the dynamic number of columns in the `SectionContent` element, and I believe we aren't anticipating any plans to do so, so there's an argument there that the current grid layout is overly-complex for our needs. In the future, should we want to put more than one child element in the `SectionContent`, I believe it would be better to do so within the `flex` at the specialized widget level rather than at the generic layout level, so we can more tightly control the styles based on whatever is going into the widget content (SVGs, images/videos, spans, more complex molecular components like the survey progress bar, etc).

I tested on various dimensions and I think it's looking better. Let me know what you think about this approach and if I missed anything. I updated the PR description with a more detailed laundry list of things I check for when I test.

ericdrosas87:
Removing the grid makes sense

ericdrosas87:
Screenshot 2026-04-06 at 1 46 35 PM

The only issue is when only one widget is enabled it isn't center aligned

PR #1091, opened 2026-05-04
fix: fix query parms for skyviewer embed

ericdrosas87:
The Skyviewer embed requires `fov` and `target` query params to focus where the embed. Previously, `ra` and `dec` were being passed in which was causing the survey to focus on it's default starting `ra` and `dec`. The solution was to concatenate `ra` and `dec` into the `target` query param.

Because the `URLSearchParams` web API will automatically encode special characters there is a `.replace()` to add back the `+`. Unfortunately the `URLSearchParams.toString()` method doesn't offer an option to not encode.

Perhaps we could modify the Skyviewer client embed page to decode URL encoded characters, but that will be addressed in a separate issue - should we decide to do so.


ericdrosas87:
FYI @soshwosh

jeffinnes (approved):

PR #1096, opened 2026-05-15
feat: added Sunrise/Sunset widget for Summit Status Dashboard

### LINKED ISSUE #1092: Add `SunTimes` widget to Summit Status Dashboard compact view
Create a new component for the `SunTimes` widget and add it to the `SummitStatusDashboardCompactView` component. Update the` SummitStatusDashboardCompactView` component to take in two new fields from Craft, `showSunTimeWidget` and `sunTimeWIdgetCaption`. There is no need to add a new data source to the Hasura action; use the existing utils/libs to compute the sunrise and sunset times, similar to the legacy proof-of-concept code in `components/widgets/CurrentData/patterns/Daylight/index.js`.

soshwosh:
Resolves #1092 

### Things to Note
**Icons**
I left TODOs for a future ticket to add the sunset and sunrise icons once they're created.

**Timezones**
The Javascript `Date` type doesn't take timezone, and doesn't offer a way to manually set timezone. It automatically assumes the user's local timezone. You're supposed to do all your date manipulation then format it at the end. The date and time values I start with come from the `SummitDataContextProvider` as strings of CST values, but when you create `Date` objects from them to then do any manipulation, they get assigned your local timezone (the actual hour/minute/day/month/year values don't change). The `getTimes()` util expects a `Date` object and attempts to cast it's return value to CST, but the `Date` object doesn't actually allow it. All the math is right and pretty accurate (sometimes a couple minutes off, but as long as it was between the window of the start of sunset/sunrise to the end of sunset/sunrise I thought it was fine). I left comments about this in the code, let me know if anything is unclear or if there is a cleaner way to do this. 

### Testing
- Pull the `sp-440-add-solarelevation-widget-for-ssd` branch on `rubin-obs-api` if it is not already merged into `develop`.
- Spin up Craft and ensure your client is pointed to the local API before spinning up your client.
- In Craft, add a add a `SummitStatusCompactView` block to a page and switch on the **Sunrise Sunset Times** lightswitch toggle.
- In the client, pull up the page with the `SummitStatusCompactView` block and confirm you see the SunriseSunsetTimes widget:
image

- Note the current time in Chile in the `LocaleContextInfoBar`:
image

- Use an online tool to look up the sunrise/sunset times at or near these coordinates: `lat = -30.244639`, `long = -70.749417`.
- Confirm that:
  - If it is before sunrise today in Chile (the current round of observing is still ongoing), the widget shows yesterday’s sunset and today’s sunrise.
  - If it is after sunrise today in Chile (aka, the last round of observing has completed), it shows today’s sunset and tomorrow’s sunrise.
- Verify responsive styling.


components/dynamic/SummitData/SunsetSunrise/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const yesterdaysTimes = getTimes(yesterday, lat, long, altitude);
+    sunset = yesterdaysTimes.sunset;
+    sunrise = todaysTimes.sunrise;
+  } else {
+    const tomorrow = new Date(currentCSTDateTime);
+    tomorrow.setDate(tomorrow.getDate() + 1);
+    const tomorrowsTimes = getTimes(tomorrow, lat, long, altitude);
+    sunset = todaysTimes.sunset;
+    sunrise = tomorrowsTimes.sunrise;
+  }
```
jeffinnes, line 46: **👏 praise:** ‏That is a doozy of a brain twister. Nice work sorting it out.

components/dynamic/SummitData/SunsetSunrise/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+): { sunset: Date; sunrise: Date } {
+  const [year, month, day] = currentCSTDateStr.split("-").map(Number);
+  const [time, modifier] = currentCSTTimeStr.split(" ");
+  const [hours, minutes] = time.split(":").map(Number);
+
+  // Note that the year/month/day/hour/minute values from the context provider are in CST,
+  // but the Date() object holds no timezone information and will treat the input as local time.
+  const currentCSTDateTime = new Date(year, month - 1, day, hours, minutes);
+
+  // The returned Date from getTimes() are also CST values, but will have the local timezone appended if you print it.
```
jeffinnes, line 29: **💭 thought:** ‏This is an interesting way to deal with the timezones. Like you noted, it coerces the local date to the CST date but with the local timezone. This _fine_ we just have to remember that we can never do date math with it because the underlying UTC time in the date object is wrong.

One day maybe we can reimplement this library using the Temporal API when it is supported in Safari (or sooner if we are okay with using the [polyfill](https://github.com/js-temporal/temporal-polyfill)).

jeffinnes (approved):

jeffinnes (approved):
Confirmed the translation works and the time is now shown in 24 hour format.

Issue #1093, opened 2026-05-12
Update link preview metadata for Rubin Basics


ericdrosas87:
It looks like the link previews for the Rubin Basics links aren't rendering correctly. Specifically the image used is the default image. This needs to be fixed so that the header image is used in place of the default image.

PR #1103, opened 2026-05-27
fix: updated Open Graph config for Rubin Basics

ericdrosas87:
Next.js can be configured to provide Open Graph the data necessary to show text or images as a link preview, among other things. Next.js exposes a function called [generateMetadata()](https://nextjs.org/docs/app/api-reference/functions/generate-metadata) that is resolved during page render and so we put it at the `layout.tsx` levels. The layouts rely on a GraphQL query to the Craft API to get the metadata for Open graph and Rubin Basics was missing from the query.

Also note that I added a subfolder `README.md` and updated an existing `README.md` as we should always be mindful to do.


ericdrosas87:
Resolves #1093

ericdrosas87:
FYI @jeffinnes @soshwosh

PR #1105, opened 2026-06-02
feat: create gallery spotlight

### LINKED ISSUE #1098: Update observatory location for SSD Sunrise/Sunset calculations
Update the hardcoded observatory location values in `lib/observatory.js` to match the key values listed [here](https://rubinobservatory.org/for-scientists/rubin-101/key-numbers) (also listed below for posterity) in order to get more accurate sunset and sunrise time calculations for the Sunset/Sunrise Times widget in the Summit Status Dashboard. 

Site coordinates: latitude -30:14:40.68, longitude -70:44:57.90
Altitude: 2647 m

### LINKED ISSUE #1101: Create generic GallerySpotlight component
Create a new component in `components/molecules` called `GallerySpotlight` to display images from any gallery that already exists on the main site. The main functionalities of this component are:
- Rotate through and display images from a provided gallery in a minimized format.
- When the image is clicked, redirect the user to the image’s individual page in the gallery in a new tab. 

The props this component should have are:
- `images`: An array of image data from whichever gallery is provided in Craft. The data should include image ids, Canto urls, captions, alternative text, and photo credits.  
- `width`, `height`: Size dimensions, expected to be overridden when this component is used in specific contexts. 
- `displayImagesRandomly`: A flag for setting random vs ordered image selection  
- `frequency`: Frequency for how often to change the image. Add the prop but only implement some default functionality to cycle through the images in the order that that are provided. Randomization can be added later.

The styling of the widget should just include sizing the images according to the dimensions provided as props. Choose the default dimensions using mobile-first design.

### LINKED ISSUE #1102: Create `WildlifeGallerySpotlight` widget
Create a new component in `components/dynamic/SummitData` called `WildlifeGallerySpotlight` to display images from a soon-to-be-added Wildlife Spotlight gallery on the main site. 

Update the `SummitStatusCompactView` GraphQL fragment in `services/craft/entries/content-blocks` to query Craft for the `showWildlifeGallerySpotlight` lightswitch, the provided gallery, and its images. Regenerate types.

Update the `SummitStatusCompactView` component to render the new `WildlifeGallerySpotlight` component based on the lightswitch toggle state. 

Pass the gallery images to `WildlifeGallerySpotlight`, which will internally render them using the generic `GallerySpotlight` component. 

It should follow the styling of the other Summit Status Dashboard widgets and provide a title and an info icon. The info icon should contain the following information: 
* A general explanation of the widget’s purpose, set in Craft in the `wildlifeGallerySpotlightCaption` field. 
* A caption describing the content of the image, set in Canto and accessed via the GQL query to Craft
* Photo credits, set in Canto and accessed via the GQL query to Craft

soshwosh:
Resolves #1101
Resolves #1102
Resolves #1098 

Changes include the addition of two new components: `GallerySpotlight` and `WildlifeGallerySpotlight`. To test:

1. Spin up the API with branch `sp-442-add-wildlife-gallery-widget-fields` (or `develop` if that branch has been merged).
2. Add the Summit Status Dashboard to a page.
3. Enable the `showWildlifeGallery` toggle and select a gallery entry.
4. Spin up the client. Confirm that the Wildlife Gallery widget shows up. 
5. Test styling (image size and containment) across various dimensions.

Testing the image flipping and synchronization:
1. In `components/molecules/GallerySpotlight/index.tsx`, on line 31, change `MS_PER_DAY` to `MS_PER_SEC` or `MS_PER_5_SEC` to get the image to change every 1 or 5 seconds. 
2. Open the client in an incognito window.
3. Confirm the images stay mostly synchronized (a couple milliseconds drift is expected/acceptable) (I think. Feel free to weigh in here).

I'd appreciate a close look at my use of `setInterval` in the `useEffect` hook in the `GallerySpotlight` component, and how I'm forcing the rerender based on the `currentDate` prop, to ensure the logic is safe and there won't be any unintended effects due to rerenders and whatnot. I think I covered all my bases, but a deep double check never hurts!


lib/i18n/localeStrings/en/translation.json
```diff
@@ -483,6 +483,9 @@
       },
       "sunrise_sunset" : {
         "title": "Sunset and Sunrise"
+      },
+      "wildlife_spotlight" : {
+        "title": "Wildlife Spotlight"
```
jeffinnes, line 488: **❔ question:** ‏Is the Spanish translation going to be added as part of this PR or picked up in a separate one?

components/dynamic/SummitData/WildlifeGallerySpotlight/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  tooltipText,
+  gallery,
+}) => {
+  const { t } = useTranslation();
+
+  // If bad data: show the title, offline icon, offline message, and the info icon if applicable
+  if (gallery === undefined || gallery === null) {
+    return (
+      <WidgetSection
+        tooltipText={tooltipText}
```
jeffinnes, line 24: **💭 thought:** ‏There is a type mismatch were this component defines `tooltipText` as `string | null` but the `WidgetSection` component is expecting it to be `string | undefined`. Because this seems to be repeated in other SSD widgets and not causing any actual issues, I don't think it's worth addressing in this PR. It's probably worth backlogging an issue so we can come back to it someday™.

components/molecules/GallerySpotlight/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const [currentDate, setCurrentDate] = useState(Date.now());
+  const galleryStr = "/gallery/collections/" + gallerySlug + "/";
+  const startDate = new Date("2026-01-01T00:00:00.000Z").getTime();
+
+  useEffect(() => {
+    const intervalId = setInterval(() => {
+      setCurrentDate(Date.now());
+    }, imageSwapInterval);
+    return () => clearInterval(intervalId);
+  }, [imageSwapInterval]);
```
jeffinnes, line 44: **👏 praise:** ‏This is good implementation of using `useEffect` to re-render the component on the desired interval. The `clearInterval` will prevent a memory leak and I don't see anything that would cause unnecessary re-renders.

jeffinnes (approved):

lib/i18n/localeStrings/en/translation.json
soshwosh: Good catch, I'll get that into this PR!

components/dynamic/SummitData/WildlifeGallerySpotlight/index.tsx
soshwosh: I'll actually fix this real quick and tack it onto this PR! It's small and easy to test while I have everything set up.

lib/i18n/localeStrings/en/translation.json
soshwosh: Plot twist, it's taking a while to get the translations and I need some of the GQL changes for another task, so i'll open a new PR for this.

PR #1117, opened 2026-06-15
fix: added ADA improvements to Rubin Dash

### LINKED ISSUE #1113: Implement accessibility improvements to the Rubin DASH
Based off the [Rubin DASH accessibility testing](https://github.com/lsst-epo/rubin-obs-client/issues/1109) results, implement the following to improve the navigability and usability of the Rubin DASH via screenreader.

- [x] Add alt text to the time in the Locale Info Bar to say "Locale Time" before reading the hours and minutes.
- [x] Add alt text to the Survey Progress number and percentage to replace the current voiceover, "0%, 0% , Progress Indicator", with something like "x percent" or "x percent progress".
- [x] Add alt text to to the Sunset/Sunrise Times widget to indicate which time is which.
- [x] Add an aria label to the Wildlife Spotlight images.
- [x] Add some auditory indication that the image in the Wildlife Spotlight is clickable.
- [x] Fix alt text from Canto not being read aloud.
- [x] Fix tooltip navigation and voiceover content.

The tooltip fixes will require some troubleshooting/discovery.

soshwosh:
Resolves #1113

Adds some ARIA attribute for better navigation of the Rubin Dash by a screenreader.

To test:
1. Open up the screenreader on your Mac: Cmd + Fn + F5
2. Double click on an element in the Locale Context Info bar.
3. Navigate through the bar, title, widgets, and caption by clicking Ctrl + Option + > and verify the following:
- [x] In the Locale Info Bar, the time reads "Locale Time" before the hours and minutes.
- [x] In the Survey Progress widget, the number reads "x percent progress made".
- [x] In the Sunset/Sunrise Times widget, it reads either "Sunset" or "Sunrise" before the time.
- [x] In the Wildlife Spotlight images, the image title is read aloud. No raw link is read.
- [x] In the tooltips, the tooltip text is read aloud when focused.

Please review the styling on various screen dimensions as well.


components/molecules/LocaleContextInfoBar/styles.module.css
```diff
@@ -1,3 +1,17 @@
+.srOnly {
+  position: absolute;
+  /* left:-10000px; */
```
jeffinnes: **🤓 nitpick:** ‏This commented style rule should be deleted here, and in the other styles files, if it is not needed.

components/molecules/GallerySpotlight/index.tsx
```diff
@@ -81,7 +81,7 @@ const GallerySpotlight: FC<GallerySpotlightProps> = ({
           src={imageUrl}
           width={width}
           height={height}
-          alt={altTextEn}
+          alt={titleEn || "wildlife image"}
```
jeffinnes: **💭 thought:** ‏Since this component is a generally reusable gallery spotlight, I think this fallback alt text should be something more generalized but vaguely descriptive (Ideally alt text should describe what's in the picture, but obviously we don't know what will be there). Maybe something like "Photo gallery highlighted image"?

components/layout/SummitStatus/WidgetSection/index.js
```diff
@@ -61,6 +63,8 @@ WidgetSection.propTypes = {
   isOpen: PropTypes.bool,
   isOffline: PropTypes.bool,
   onToggleCallback: PropTypes.func,
+  tooltipId: PropTypes.string,
```
jeffinnes: **❔ question:** ‏Is `tooltipId` meant to be implemented in this component? It doesn't look like it's deconstructed from the props here, but I do see that it's being passed to the component in `components/dynamic/SummitData/WeatherCondition/index.tsx`. It also looks like the `WeatherCondition`   implementation is the only one passing a `tooltipId`.

components/atomic/Tooltip/index.tsx
```diff
@@ -3,21 +3,37 @@ import styles from "./styles.module.css";
...
-      </div>
-    );
-  }
+const Tooltip: FC<TooltipProps> = ({
+  value,
+  isRendered,
+  isVisible,
+  tooltipFormatter,
+}) => {
+  const showVisually = isRendered && isVisible;
```
jeffinnes, line 17: **👏 praise:** ‏Good call evaluating this here to keep the ternary cleaner!

jeffinnes (changes requested):
Good work digging into how to test with screenreaders and implementing best practices. I also appreciated the instructions on how to use the screenreader.

components/molecules/GallerySpotlight/index.tsx
soshwosh: Good catch, will do!

components/layout/SummitStatus/WidgetSection/index.js
soshwosh: I was using this at one point when I was trying `aria-describedby` for the tooltip announce, but it's no longer needed and I missed this in cleanup. I'll remove it!

jeffinnes (approved):

PR #1118, opened 2026-06-16
feat: removed hasura metadata

ericdrosas87:
We need to remove this header so the proxy works as expected


ericdrosas87:
FYI @jeffinnes @soshwosh

